PR #30, opened 2015-01-28
start/end date in datetime not timestamp format 

raul72:
start/end date in datetime not timestamp format for Time_Tracking::getSummaryByUser

otherwise SQL error:
Warning (Code 1292): Incorrect datetime value: '1422223200' for column 'ttr_created_date' at row 1


glensc:
Could use DateTime object between functions, i.e getSummaryByUser

PR #31, opened 2015-01-29
Restore $extra_title functionality

balsdorf:
Also always append APP_TITLE to title


glensc:
the old templates actually used TOOL_CAPTION not APP_TITLE

PR #40, opened 2015-03-18
Fix error parsing confusing email addresses.

balsdorf:
Added tests for parsing email addresses.

This is the first time I've added a test case I believe so I wanted you to check it out and let me know if there are any issues.


glensc:
looks good. i will refactor it to use `@dataProvider`

PR #45, opened 2015-04-06
Added a display of if an individual file has been uploaded successfully

balsdorf:
This solves a problem where a large / slow moving upload is not successfully uploaded before the parent form is submitted.


glensc:
yep. and eventually should get rid of `$smarty.get.legacy` as well. it's not possible to use it in all places anyway (like send email popup)

PR #60, opened 2015-06-30
Chosen (Assignee selection dropdown) fixes

slay123:
Global:
* don't create bigger area, when clicking on Choose Assignees, make it 27px, let the dropdown act like dropdown not new 240px height div
* style changes: separate default chosen dropdowns and bulk update dropdown. nicer dropdown shadows 

Bulk update:
* when bulk update, add dummy/empty option, so first assignee will not be selected automatically when clicking on dropdown
* bulk update vs chosen conflict - dropdown makes page longer and chosen selects the item under mouse, lets open assignee list to the top, not below
* minor style changes: cast shadow to top, not to bottom, as dropdown opens to top

### DIFF
--- a/htdocs/css/page.css
+++ b/htdocs/css/page.css
@@ -360,7 +360,11 @@ body#adv_search #date_fields, body#adv_search #custom_fields_row {
     height: 27px;
 }
 .chosen-container .chosen-drop {
+    top: 0;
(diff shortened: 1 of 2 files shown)


slay123:
Update in issue:
screenshot 2015-06-30 10 35 19

Bulk update:
screenshot 2015-06-30 10 33 36

PR #61, opened 2015-07-01
CAS Auth Backend

balsdorf:
This adds support for Central Authentication Service (CAS) in Eventum. The backend totally replaces the built in auth and does not allow fallback. If configured, Eventum will create new users.

Changes to auth interface are:
* Addition of getExternalLoginURL() method. If set no login form will be displayed. Instead the user
 will be redirect to an external URL to authenticate.
* Addition of checkAuthentication() method. This is called before the normal Eventum authentication
process.
* Addition of logout() method. This is called after the Eventum logout process is completed.

I'm not 100% sure that getExternalLoginURL() and checkAuthentication() is the cleanest way but I haven't thought of a better way.

Please let me know any feedback you have on this or if you have any better ideas on how to handle systems with external authentication like this.

### DIFF
--- a/docs/examples/config/cas.php
+++ b/docs/examples/config/cas.php
@@ -1,5 +1,5 @@
 <?php
-$cas_setup = array(
+return array(
     'host'  =>  'localhost',
     'port'  =>  443,
     'context' =>  '/cas',
--- a/docs/examples/config/cas.php
+++ b/docs/examples/config/cas.php
@@ -1,0 +1,12 @@
+<?php
+$cas_setup = array(
+    'host'  =>  'localhost',
+    'port'  =>  443,
+    'context' =>  '/cas',
+    'customer_id_attribute' => '',
+    'contact_id_attribute' => '',
+    'create_users' => true,
+    'default_role' => array(
+        1 =>    4,
+    ),
+);
--- a/lib/eventum/auth/class.cas_auth_backend.php
+++ b/lib/eventum/auth/class.cas_auth_backend.php
@@ -29,7 +29,7 @@
 /**
  * This auth backend integrates with a CAS server
  *
- * This backend will look for users in the default mysql backend if no LDAP
+ * This backend will look for users in the default mysql backend if no CAS
  * user is found. This behaviour may be configurable in the future.
  *
  * Set define('APP_AUTH_BACKEND', 'CAS_Auth_Backend') in the config file and
@@ -102,7 +102,7 @@ class CAS_Auth_Backend implements Auth_Backend_Interface
             // do not reset user password, it maybe be set locally before this
             unset($data['password']);
 
-            // perspective what is main address and what is alias may be different in ldap and in eventum
+            // perspective what is main address and what is alias may be different in CAS and in eventum
             $emails = array($remote['mail']);
             $email = User::getEmail($usr_id);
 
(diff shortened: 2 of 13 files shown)


docs/examples/config/cas.php
```diff
@@ -0,0 +1,12 @@
+<?php
+$cas_setup = array(
+    'host'  =>  'localhost',
```
glensc: i would prefer if you just return the array and in callee use `$cas_setup = require 'config.php';`

lib/eventum/auth/class.cas_auth_backend.php
```diff
@@ -0,0 +1,356 @@
...
+// | Authors: Bryan Alsdorf <balsdorf@gmail.com>                          |
+// | Authors: Elan Ruusamäe <glen@delfi.ee>                               |
+// +----------------------------------------------------------------------+
+
+/**
+ * This auth backend integrates with a CAS server
+ *
+ * This backend will look for users in the default mysql backend if no LDAP
+ * user is found. This behaviour may be configurable in the future.
+ *
```
glensc: the class comment needs updating

lib/eventum/auth/class.cas_auth_backend.php
```diff
@@ -0,0 +1,356 @@
...
+            $data['contact_id'] = $remote[$setup['contact_id_attribute']];
+        }
+
+        // if local user found, update it and return usr id
+        if ($usr_id) {
+            // do not reset user password, it maybe be set locally before this
+            unset($data['password']);
+
+            // perspective what is main address and what is alias may be different in ldap and in eventum
+            $emails = array($remote['mail']);
```
glensc: comment needs updating (this is not ldap backend)

lib/eventum/auth/class.cas_auth_backend.php
```diff
@@ -0,0 +1,356 @@
...
+        foreach ($aliases as $alias) {
+            User::addAlias($usr_id, $alias);
+        }
+    }
+
+
+    /**
+     * With CAS we cannot do a simple password check like this. This will prevent the CLI from working
+     * so at some point in the future we need to find a solution.
+     *
```
glensc: probably we should setup per user api keys in user preferences, and use such api token to access cli instead of password auth

lib/eventum/class.session.php
```diff
@@ -108,8 +108,6 @@ public static function init($usr_id)
      */
     public static function verify($usr_id)
     {
-        session_start();
-
         // Don't check the IP of the session, since this caused problems for users that use a proxy farm that uses
```
glensc: any reason why you removed session_start?

glensc:
looks good overall, but could you rebase against origin/master (to remove the merge commits) before making further changes?

lib/eventum/class.session.php
balsdorf: It was causing a warning if any other code initiated a session before it. My testing showed that it was not needed as Session::init() (which is called by Session::verify() ) checks if a session is active and if not activates it. However, in I will add a similar check here in case something I couldn't find was depending on it.

docs/examples/config/cas.php
balsdorf: I did not know about that trick. Thanks!

lib/eventum/auth/class.cas_auth_backend.php
balsdorf: Agreed that this would be the best way to go. When I implement CAS internally I will either have to setup auth keys or some workaround for this.

PR #64, opened 2015-07-15
add maxwidth to issue description, so it will be scrollbar, when too …

slay123:
…many chars on same line without whitespace
before:
screenshot 2015-07-15 15 48 59

after:
screenshot 2015-07-15 15 49 41

### DIFF
--- a/htdocs/css/main.css
+++ b/htdocs/css/main.css
@@ -374,6 +374,13 @@ table.grid tr.no_results td {
     text-align: center;
     font-style: italic;
 }
+/* if issue description is too wide */
+#issue_description {
+    display: inline-block;
+/* 82% from viewport width */
+    max-width: 82vw;
+    overflow: auto;
+}
 
 /* End Grid */
 


htdocs/css/main.css
```diff
@@ -374,6 +374,13 @@ table.grid tr.no_results td {
     text-align: center;
     font-style: italic;
 }
+/* if issue description is too wide */
+#issue_description {
+    display: inline-block;
+/* 82% from viewport width */
+    max-width: 82vw;
```
glensc: i wonder what's browser compatibility for this css unit?

slay123: According to:
https://css-tricks.com/viewport-sized-typography/ and
http://caniuse.com/#feat=viewport-units

IE 10+, Firefox 19+, Chrome 34+, Safari 7+, Android 4.4+, iOS 6+ - Supported

For IE should probably add some IE specific CSS hack, something like:
max-width: 995px\9;

Elan Ruusamäe wrote on 15.07.2015 16:59:

PR #67, opened 2015-08-14
Add admin interface for required fields.

balsdorf:
The "Fields to Display" manage page now allows you to select which fields should be required.
The "Allow Unassigned issues" setting in admin is now removed.

### DIFF
--- a/htdocs/js/page.js
+++ b/htdocs/js/page.js
@@ -822,6 +822,11 @@ new_issue.validateForm = function()
         expected_res_date_field.data('required')) {
         Validation.errors[Validation.errors.length] = new Option('Expected Resolution Date', 'expected_resolution_date');
     }
+    var associated_issues_field = Eventum.getField('associated_issues')
+    if (associated_issues_field.attr('type') != 'hidden' && associated_issues_field.val() == '' &&
+        associated_issues_field.data('required')) {
+        Validation.errors[Validation.errors.length] = new Option('Associated Issues', 'associated_issues_field');
+    }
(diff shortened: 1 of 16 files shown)


glensc:
Do I get this right, you dropped `allow_unassigned_issues` option? Why?

In our workflow issues created via mail are not assigned, each department reviews unassigned issues list and assigns it if that issue is to him/her or someone in the team.

balsdorf:
I dropped it because it conflicted if you marked the "assigned" field as
required.

Your workflow still works perfectly.

The only workflow that is slightly different is if someone does not want
unassigned issues. It will still require an assignee on the create page but
if you go to the update page you can remove all assignees. I didn't
consider this a problem. If it is the best solution would be to make update
honor the "required" options from admin.

On Mon, Aug 17, 2015 at 3:21 AM, Elan Ruusamäe <notifications@github.com>
wrote:

> Do I get this right, you dropped allow_unassigned_issues option? Why?
>
> In our workflow issues created via mail are not assigned, each department
> reviews unassigned issues list and assigns it if that issue is to him/her
> or someone in the team.
>
> —
> <https://github.com/eventum/eventum/pull/67#issuecomment-131727761>.
>



-- 
Bryan Alsdorf

glensc:
can you make sql patch to make required fields such that they were before this pull-request?

as if i go to admin or new issue page, none of the fields are required, but i'm pretty sure category and priority were required before.

otherwise if can't do that, write something to UPGRADE_NOTES file in git

Issue #66, opened 2015-08-05
Download emails in 3.0.2 after upgrade from 2.3.6


cpinfold:
download_emails.php calls Email_Account::getAccountID (on line 172) and it returns 0 after an upgrade from 2.3.6.  DB_Helper::getInstance()->getOne() is returning null within that function.  Executing the SQL from getAccountID returns the correct result when executed manually. (See comment 4 for clarification)

download_emails.php was called with only a user and host (not folder) as my accounts are using POP3.

I'll try and trace it further, for now I have hard-coded the account ID's for my accounts.  After hard-coding, the balance of the script is working correctly.

Also the UPGRADE doc should be updated to point out the cron scripts moved from the crons folder to the bin folder.

plindry:
I have the same problem  in line near 169 "$account_id = Email_Account::getAccountID($config['username'], $config['hostname'], $config['mailbox']);" it always return 0. I've hardcoded the id in the database for the only one email configured.

glensc:
care to make patch (as pull-request) with proper fix?

cpinfold:
I've found the issue and can write something up, any preference as to the solution:

The issue is caused when using POP accounts, there is no folder.  Because $config['mailbox'] is null, however the database contains '' (probably a legacy issue)

Eg.
select * from eventum_email_account where ema_folder is null; Returns nothing.
select * from eventum_email_account where ema_folder = ''; Returns the email account.

Quick fix for affected users:  If you go and edit each of the email account settings and save them, the current page stores NULL (this fixes the issue)

Options:
1) Replace '' with NULL on upgrade as a DB change?
2) Cause null folder search to also allow '', eg *or ema_folder = ''* 
3) Change query on null folder to omit ema_folder from search
4) Do nothing and require users to manually fix

Other ideas?

glensc:
i think safest is to change sql query. ping @balsdorf

balsdorf:
I'd go with #3. If $mailbox is null, don't include it in the where clause.

On Mon, Oct 5, 2015 at 11:37 AM, Elan Ruusamäe <notifications@github.com>
wrote:

> i think safest is to change sql query. ping @balsdorf
> <https://github.com/balsdorf>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/eventum/eventum/issues/66#issuecomment-145593440>.
>



-- 
Bryan Alsdorf
MariaDB Corporation | http://mariadb.com

glensc:
[...]

PR #69, opened 2015-10-06
Fix POP3 email download bug (Issue #66)

cpinfold:
Fix POP3 email download bug (Issue #66) where $mailbox was null, but a null string in the pre 3.02 database caused the email account ID to be zero.

Also changed UPGRADE.md to reflect location change of cron jobs.


glensc:
merged, btw, you had bug that you left in the original query. fixed that

PR #70, opened 2015-10-09
Add autosize plugin to all TEXAREAs

glensc:
[Autosize](http://www.jacklmoore.com/autosize/) is a small, stand-alone script to automatically adjust textarea height to fit text.


glensc:
maybe want to set [min-height or max-height](http://www.jacklmoore.com/autosize/#faq-min-max) in some places

PR #77, opened 2015-10-16
refactor password handling

glensc:
uses [password_hash](https://wiki.php.net/rfc/password_hash) family functions by default.
[compatibility library](https://github.com/ircmaxell/password_compat) for older php versions is included

new passwords are set using new algorithm, existing hashes will be verified using old methods.
auto migration of hash on successful login and hash upgrade can be added in later releases.
 
also added old password prompt when changing password and trivial check such as new and old passwords differ (auth backend said `false` anyway for such password change)

also includes sql patch to clear password for accounts where empty password was stored for whatever reason.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,6 +51,7 @@
 		"ext-session": "*",
 		"ext-spl": "*",
 		"fonts/liberation":"*",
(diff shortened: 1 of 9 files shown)


glensc:
this is ready for merge

PR #81, opened 2015-10-24
add /var path for writable data

glensc:
add `/var` path for writable data

this simplifies a lot permission setup.
inspired by symfony3 new layout.

Old Value  | New Value | Description
------------- | ------------- | -------------
`templates_c`  | `var/cache` | templates cache
`locks`  | `var/lock` | various lock and pid files
`logs`  | `var/log` | directory for logs

- [x] make the changes
- [x] update installation documents: https://github.com/eventum/eventum/wiki/Upgrading
- [x] prepare release note 

users wishing to use old directories, should define in their `config/config.php`:

```php
define('APP_TPL_COMPILE_PATH', APP_PATH . '/templates_c');
define('APP_LOCKS_PATH', APP_PATH . '/locks');
define('APP_LOG_PATH', APP_PATH . '/logs');
```


glensc:
i wonder what to write to release notes?

balsdorf:
Another excellent idea. 

For release notes, I'd put something like:

To simplify setup and directory layout we have moved all directories that contain files to which Eventum writes data during the course of its operation into /var. You may have grant write permissions on /var to your webserver.

glensc:
i do not see anything regarding file permissions in wiki

https://github.com/eventum/eventum/wiki/System-Admin%3A-Doing-a-fresh-install

i think only place where permissions are mentioned is install wizard.

glensc:
i wonder, should we make a patch that would migrate pid files from old location to new one?
logs and template cache are likely irrelevant.

balsdorf:
The only long running process I can think of is the IRC bot and it needs to
be restarted to pick up the new changes so it should move its own PID,
right?

On Mon, Oct 26, 2015 at 7:51 AM, Elan Ruusamäe <notifications@github.com>
wrote:

> i wonder, should we make a patch that would migrate pid files from old
> location to new one?
> logs and template cache are likely irrelevant.
>
> —
> <https://github.com/eventum/eventum/pull/81#issuecomment-151123407>.
>



-- 
Bryan Alsdorf

glensc:
yes, irc bot.
and also issue edit locks (#4)

issue edit locks are likely not an issue, but irc bot being started again probably is.

balsdorf:
We can just add a note that they should stop the bot before upgrading but
who knows if they actually will. Lets go ahead and move the pid files so
duplicates aren't started.  I have been trying to think of a reason not too
but I haven't found any

On Mon, Oct 26, 2015 at 8:20 AM, Elan Ruusamäe <notifications@github.com>
wrote:

> yes, irc bot.
> and also issue edit locks (#4 <https://github.com/eventum/eventum/pull/4>)
>
> issue edit locks are likely not an issue, but irc bot being started again
> probably is.
>
> —
> <https://github.com/eventum/eventum/pull/81#issuecomment-151131000>.
>



-- 
Bryan Alsdorf

glensc:
reasons could be like upgrade done with insufficient permissions. but they will see the error and retry, i hope

PR #82, opened 2015-10-25
irc bot improvements

glensc:
refactored code to be in classes and using modern php version (php5),
program moved to `bin/` dir as well.

most useful change is probably automatic pid file cleanup on daemon termination and ctrl+c


glensc:
there's some unclear TODO:

https://github.com/glensc/eventum/blob/irc-bot/lib/eventum/irc/Eventum_Bot.php#L69

what is old, what is new? what is the expected structure?

as i tested with such config, and it seemed to work:
```php
$irc_channels = array(
    'Default Project' => '#eventum',
);
```

balsdorf:
I think that TODO was to re-write the old config file into the "new" style, i.e. with "categories". However I don't think this is really needed. 99% of people will just use the old style as it is simpler. The "new" style is just for me mainly.

glensc:
but the old config already works? so the todo is complete?

confused

balsdorf:
The TODO was never completed, but it is actually not needed. It was to
update the contents of config/irc_config.php instead of having the bot
understand both formats.

However, I think understanding both formats is the better choice (what we
already have). So all that needs to be done is remove that TODO.

On Mon, Oct 26, 2015 at 8:39 AM, Elan Ruusamäe <notifications@github.com>
wrote:

> but the old config already works? so the todo is complete?
>
> confused
>
> —
> <https://github.com/eventum/eventum/pull/82#issuecomment-151135866>.
>



-- 
Bryan Alsdorf

PR #85, opened 2015-10-28
cleanup authors block and update contributing

glensc:
cleanup authors block and update contributing. mostly to ease maintaining the releases.


glensc:
@balsdorf anything else?

balsdorf:
@glensc Looks good!

PR #88, opened 2015-11-02
Update custom fields from update issue page

balsdorf:
Custom fields are now included on the update issue page. During the implementation I fixed a bunch of little custom field related bugs.

* Include custom fields in update issue page.
* Fixed bug updating custom field definitions.
* Display admin only custom fields in 'internal' color.
* Fix bug unchecking checkbox custom fields.
* Include custom field changes in issue updated email.
* Fix data leakage problem with admin only custom fields by restricting history entries to correct role.
* Added minimum role to each history entry. If no role is specified it will default to role from history type.

Custom fields are displayed in a separate box, if we overhaul the UI we might want to include them with the rest of the issue details.

### DIFF
--- a/htdocs/close.php
+++ b/htdocs/close.php
@@ -60,7 +60,7 @@ $tpl->assign('notification_list_internal', $notification_list_internal['all']);
 
 $cat = isset($_REQUEST['cat']) ? (string) $_REQUEST['cat'] : null;
 if ($cat == 'close') {
-    Custom_Field::updateValues();
+    Custom_Field::updateFromPost();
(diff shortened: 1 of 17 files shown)


glensc:
i'm going to merge this, as it's best way to test it if it's merged

glensc:
One UI thing bothers me:

Could the custom fields moved more up, i.e after the TEXTAREA, so Private and Trigger Reminders stay last ones.

The private and trigger reminders are rarely used.

balsdorf:
Done.

glensc:
well, you have to convince @balsdorf to do it then :)

PR #93, opened 2015-11-05
upgrade password hash on successful login

glensc:
the idea was to do this at least one release after new hash introduction. so here it is.

it will also upgrade new hashes if there is more secure algorithm enabled by default.


balsdorf:
This looks good to me.

PR #95, opened 2015-11-06
Issue 91 notice and IMAP missing error change

cpinfold:
@glensc I didn't see that you also fixed this until after.. this pull also fixes the notice.  I used Misc::ifSet that you referenced in the notes.

I also added a new line to the error text when imap_open is missing:
"It is installed by default on some systems but must also be enabled. See INSTALL.md"

The error previously only referenced checking the PHP docs.. the PHP site is silent on enabling.

INSTALL.md isn't there any more (so i didn't edit) - we need a note that imap needs to be installed **and** enabled.  `php5enmod -s ALL imap` should do it on most linux machines.  Works on Ubuntu.

### DIFF
--- a/bin/download_emails.php
+++ b/bin/download_emails.php
@@ -131,8 +131,8 @@ function getParams()
 if (!function_exists('imap_open')) {
     fatal(
         'Eventum requires the IMAP extension in order to download messages saved on a IMAP/POP3 mailbox.',
-        'It is installed by default on some systems but must also be enabled. See INSTALL.md',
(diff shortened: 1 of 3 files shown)


glensc:
documentation is moved to wiki, but as wiki is in restructuring, i'm not certain where to link to

glensc:
please rebase this PR against master, and push -f to update github

cpinfold:
@glensc I rebased and changed the error messages.

PR #96, opened 2015-11-06
Ajax email test

cpinfold:
Display of result on page instead of popup.  Popup was not a form in this case.

Also changed base.tpl to move full_width_messages div outside block so it appears on every page.
This is in preparation for later enhancing other pages to use this div
for AJAX response messages. base_full.tpl was modified and the messages block was
removed so as not to override base.tpl (looked duplicative anyway)

I am proposing this as a template for making similar changes to other pages which use pop-ups only for results.  Any clean-up/enhancements (particularly css) are welcome.

### DIFF
--- a/templates/get_emails_ajax.tpl.html
+++ b/templates/get_emails_ajax.tpl.html
@@ -1,0 +1,14 @@
+{* AJAX response message *}
+<span class="default">
+
+  {if $error == "imap_extension_missing"}
+    <b>Error: Eventum requires the IMAP extension in order to connect to IMAP/POP3 servers.<br /><br />
+    Please refer to the PHP manual for more details about how to enable the IMAP extension.</b>
+  {elseif $error == "hostname_resolv_error"}
+    <b>The provided hostname could not be resolved. Please check your information and try again.</b>
+  {elseif $error == "could_not_connect"}
+    <b>Could not connect to the server with the provided information.</b>
+  {else}
+    <b>Thank you, the connection to the email server was created successfully.</b>
+  {/if}
+</span>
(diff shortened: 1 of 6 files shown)


templates/get_emails_ajax.tpl.html
```diff
@@ -0,0 +1,14 @@
+{* AJAX response message *}
```
glensc: perhaps just use subdir for ajax type templates? say 'ajax' ?

templates/manage/email_accounts.tpl.html
```diff
@@ -91,6 +79,18 @@
       toggleFolderField();
       $('#email_account_form').submit(validateForm);
       $('.select_all').click(function() { Eventum.toggleCheckAll('items[]'); });
+      $('#testSettings').click(function(event) {
+           event.preventDefault();
+           $('#full_width_messages').removeClass().html('Please wait.. testing connection.').addClass('info');
+           var serializedData = $('#email_account_form').serialize();
+           $.ajax({
+               type: "POST",
+               url: "check_email_settings.php",
```
glensc: i think we should create common entry point for ajax calls. there are too many of them, and each one does things by it's own, not sure which is even correct one. i.e how to properly check for authenticated user and role.

see here:
https://github.com/eventum/eventum/tree/v3.0.5/htdocs/ajax

they all do the authenticity check differently.

cpinfold: I agree a shared JS function would probably be better.  For authentication,
I have always used the same authentication on the ajax PHP files as
others.  The page still has access to the session, do we need to do
anything differently than for example list.php?

I see what you mean in the ajax/ sub directory.  Any reason we couldn't add
`Auth::checkAuthentication();` to the bottom of init.php and avoid calling
it in every other file? We would have to skip authentication if init.php
was included from a cli call (wouldn't we?).

Let me know what you and @balsdorf think and I'll rework this example.

Craig.

glensc:
merged. can do cleanup/refactor/etc later, i.e discuss with @balsdorf. maybe i'll just do PR with idea what i had in mind.

glensc:
Now unused get_emails.tpl.html will be dropped in https://github.com/eventum/eventum/pull/937

PR #97, opened 2015-11-06
add logging framework based on monolog

glensc:
this adds framework to have common logging in code.
as we're in legacy-codebase situation, best fit was to use [Monolog\Registry](https://github.com/Seldaek/monolog/blob/1.17.2/doc/03-utilities.md)

overview is simple, call this in `init.php`:

```php
Logger::initialize();
```

and when want to log:
```php
Logger::db()->error('Sent to $db Logger instance');
Logger::app()->error('Sent to $app Logger instance');
```

Levels below `WARNING` are not logged anywhere, but it's pretty easy to setup something to do so. it's in single location (`Logger` class).

sub tasks:
- [x] replace db error logging
- [ ] add logging to auth actions
- [x] add `errors.log` logger, or just deprecate it in favour of general log?
- [x] add `login_attempts.log` logger, or add new channel 'auth'?
- [x] add mail output support

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,6 +57,7 @@
 		"fonts/liberation":"*",
 		"ircmaxell/password-compat": "~1.0.4",
 		"ircmaxell/random-lib": "~1.1.0",
+		"monolog/monolog": "~1.17.2",
 		"pear-pear.php.net/mail_mimedecode": "*",
 		"pear-pear.php.net/math_stats": "*",
 		"pear-pear.php.net/net_ldap2": "~2.1",
--- a/htdocs/ajax/upload.php
+++ b/htdocs/ajax/upload.php
@@ -55,8 +55,7 @@ try {
         'error' => $code ? $code : -1,
         'message' => $e->getMessage(),
     );
-    error_log($e->getMessage());
-    error_log($e->getTraceAsString());
+    Logger::app()->error($e);
 }
 
 header('Content-Type: application/json; charset=UTF-8');
--- a/htdocs/file_upload.php
+++ b/htdocs/file_upload.php
@@ -58,8 +58,7 @@ if ($cat == 'upload_file') {
         Attachment::attachFiles($issue_id, $usr_id, $iaf_ids, $internal_only, $file_description);
         $res = 1;
     } catch (Exception $e) {
-        error_log($e->getMessage());
-        error_log($e->getTraceAsString());
+        Logger::app()->error($e);
         $res = -1;
     }
 
--- a/htdocs/offline.php
+++ b/htdocs/offline.php
@@ -26,7 +26,10 @@
 // | Boston, MA 02110-1301, USA.                                          |
 // +----------------------------------------------------------------------+
 
-require_once __DIR__ . '/../init.php';
+// this file may be called from db_helper, so init already called
+if (!defined('APP_PATH')) {
+    require_once __DIR__ . '/../init.php';
+}
 
 $tpl = new Template_Helper();
 if (php_sapi_name() == 'cli') {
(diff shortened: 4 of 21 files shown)


glensc:
regarding mail handler. do we still need such nicely formatted mails? and does they need to be sent via eventum mail queue? or can be sent directly via `mail()` as well?

```
Hello,

An error was found on /usr/share/eventum/lib/eventum/class.support.php:402

The error message passed to us was:

'Error while connecting to the email server - Can not authenticate to IMAP server: [CLOSED] IMAP connection broken (authenticate)'

A backtrace is available:

APP_PATH/bin/download_emails.php:226
  Support::connectEmailServer(Array(13))


-- 
Sincerely yours, Automated Error_Handler Class 
```

poing:
There are a few pros and cons to changing the mail method.  

The mail handler shows the delivery status in the 'mail queue log' for each issue.  Useful for tracking and verification messages have been sent to specific users.

If you go with the direct mail method, you could eliminate sending individual messages to each member of the notification list.  But that could result in a reply-to-all nightmare.   

On the reverse side of the reply-to-all nightmare, it could solve the 'I forwarded to x-user because it said "Hello X-User", but didn't see the person CC'd in the message' problem.  :smile:


Brian

glensc:
@poing err, you're totally out of context. [this question here](https://github.com/eventum/eventum/pull/97#issuecomment-155001170) was about [error reporting](https://github.com/glensc/eventum/blob/0660a4a6407560fc291f9ff2371cbc06ea2bd6b6/lib/eventum/class.error_handler.php#L46). and the [current implementation](https://github.com/glensc/eventum/blob/0660a4a6407560fc291f9ff2371cbc06ea2bd6b6/lib/eventum/Logger.php#L112-L135) in this branch would send out one mail for each recipient separately.

poing:
Oops! :blush:

balsdorf:
I don't think we need pretty formatted error emails and it is better they are sent through mail() instead of the mail queue. I've seen cases where errors cause an avalanche of DB activity which makes things even worse.

balsdorf:
I like this PR a lot. Our logging has been rather limited and basic. This provides a centralized way to go forward. 

I think login attempts can be handled through a new "auth" channel.

Can we make anything with a level of ERROR or above go into the errors.log still and have everything else go into eventum.log? I guess if it all goes into one file you can just search to see what is critical.

glensc:
yes, that's the cool part of it, that you can configure that some handler only accepts messages above certain level and ignore lower (not sure if opposite possible too, but usually not needed).

so the idea was to log at first everything over 'app' channel, and levels above 'error' get an email handler too (or maybe use 'critical' level for that?)

what level 'db' errors should be used?

i think for simplicity we should use channels only for purposes of having different ident in logs, so that all channels really do the same:

something like:
- level >= debug -> debug.log
- level >= info -> eventum.log
- level >= error -> mail

and with all that a level==error would be logged to debug.log, eventum.log and mail.

so, i'll leave the 'mail' handler to be monolog default, we can customize  the output later if really needed.

balsdorf:
'db' should all be error I believe, at least that is how we use it now. I guess we could log warnings that are returned from MySQL but I'm not sure useful that is

glensc:
i mean what is the LEVEL for db, not CHANNEL.

glensc:
ok. @balsdorf: made my changes, so it's ready to merge. auth logging can be added later (as it  doesn't exist now either).

i want you to have closer look how we do the logging.
Error_Handler class can be dropped completely, no longer used. (will remove it a bit later)

setup part not updated, as i'm not clear which files we will log to, need your input here.

glensc:
@balsdorf ready for merge

balsdorf:
Sounds good.

Once you merge I want to try to make it work with https://getsentry.com/

On Fri, Nov 13, 2015 at 4:59 PM, Elan Ruusamäe <notifications@github.com>
wrote:

> @balsdorf <https://github.com/balsdorf> ready for merge
>
> —
> <https://github.com/eventum/eventum/pull/97#issuecomment-156582698>.
>



-- 
Bryan Alsdorf

PR #101, opened 2015-11-12
Fix static notifications in class.issue.php

cpinfold:
Minor declaration changes to avoid notifications.


glensc:
thanks! weird that these still exist. i ran some check over whole code at some point and reached to zero for those static calls on non-static method reports.

PR #104, opened 2015-11-13
add .htaccess to project root

glensc:
this will at least redirect to htdocs subdir for poortly setup
installations.

originally i intended to add there `Require all denied` and `Require all granted` when directory is `htdocs`, but wasn't able to accomplish that without the same `.htaccess` matching in `htdocs/` dir.

so just kept rewrite rules. which also originally were conditional `<IfModule mod_rewrite.c>`, but removed condition, so that if you two of these:
  - expose directory outside htdocs to web
  - don't have mod_rewrite enabled either 

you will get 500 error about misconfiguration and it's good indication that you should really check your installation and read some docs.

making the web safer place!

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,0 +1,21 @@
+#
+# You should not expose Eventum to web root.
+# Only htdocs directory should be exposed to webroot
+#
+# See installation documentation from wiki:
+# https://github.com/eventum/eventum/wiki/System-Admin%3A-Doing-a-fresh-install
+#
+
+# This is just workaround to redirect all requests outside /htdocs/ to /htdocs/.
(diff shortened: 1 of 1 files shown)


glensc:
@balsdorf ready for merge

balsdorf:
@glensc I might have actually merged this too soon. It will *always* try to redirect to /htdocs/ if apache is setup to process .htaccess files in parent directories. It also would always redirect to /htdocs/ ignoring the full request so things like css were broken.

I've committed a fix for this (ef2e562) that works for me, please confirm it works for you.

glensc:
thanks!

PR #107, opened 2015-11-16
Added ability to require custom fields on the edit form.

balsdorf:
As requested by @cpinfold 

Not sure about the management UI. I added an always checked (but disabled) check box for the edit form to be consistent. Not sure if it is confusing or helps.

image

### DIFF
--- a/upgrade/patches/41_custom_field_required.sql
+++ b/upgrade/patches/41_custom_field_required.sql
@@ -1,3 +1,3 @@
 -- Add field to allow custom fields to be marked as required when editing
 alter table {{%custom_field}}
-    add `fld_edit_form_required` tinyint(1) NOT NULL DEFAULT 0;
+    add `fld_edit_form_required` tinyint(1) NOT NULL DEFAULT 0 AFTER fld_close_form_required;
--- a/upgrade/patches/41_custom_field_required.sql
+++ b/upgrade/patches/41_custom_field_required.sql
@@ -1,0 +1,3 @@
+-- Add field to allow custom fields to be marked as required when editing
+alter table {{%custom_field}}
+    add `fld_edit_form_required` tinyint(1) NOT NULL DEFAULT 0;
--- a/htdocs/js/main.js
+++ b/htdocs/js/main.js
@@ -806,7 +806,6 @@ ExpandableCell.ready = function()
             ExpandableCell.expand(expand_type, list_id);
         } else {
             $.each(expand_type.split(","), function(index, value) {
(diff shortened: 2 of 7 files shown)


upgrade/patches/41_custom_field_required.sql
```diff
@@ -0,0 +1,3 @@
+-- Add field to allow custom fields to be marked as required when editing
+alter table {{%custom_field}}
+    add `fld_edit_form_required` tinyint(1) NOT NULL DEFAULT 0;
```
glensc: please add `AFTER` statement here, so fields in table would be ordered like:
 - primary key
 - int values
 - text values

this supposedly improves key lookup performance.
i'd like to follow this idea in eventum

also probably should also order int fields too, but not sure in what order, smallest size first or biggest size first. or doesn't matter, just be consistent?

balsdorf: I've updated the patch to insert that column at the correct location.
However, I didn't have time to re-order all the columns.

For a very small table like custom_field I don't see this as an issue. It
would be more important for larger tables (like issue_custom_field). Also,
custom_field will pretty much always be looked up by PK, not by one of the
later columns.

On Mon, Nov 16, 2015 at 4:45 PM, Elan Ruusamäe <notifications@github.com>
wrote:

> In upgrade/patches/41_custom_field_required.sql
> <https://github.com/eventum/eventum/pull/107#discussion_r44996307>:
>
> > @@ -0,0 +1,3 @@
> > +-- Add field to allow custom fields to be marked as required when editing
> > +alter table {{%custom_field}}
> > +    add `fld_edit_form_required` tinyint(1) NOT NULL DEFAULT 0;
>
> please add AFTER statement here, so fields in table would be ordered like:
>
>    - primary key
>    - int values
>    - text values
>
> this supposedly improves key lookup performance.
> i'd like to follow this idea in eventum
>
> also probably should also order int fields too, but not sure in what
> order, smallest size first or biggest size first. or doesn't matter, just
> be consistent?
>
> —
> <https://github.com/eventum/eventum/pull/107/files#r44996307>.
>



-- 
Bryan Alsdorf

glensc: yes, this is just for new fields.

i think i will recreate schema at some point, marking all applied patches applied, and there can do some reordering fun. or you think it's worth to do patch as well?

and do you have opinion how to order int fields? maybe some internal info? ;)

cpinfold:
Thanks Bryan!

Craig.

PR #110, opened 2015-11-25
Improve select project page and increase project cookie lifespan.

balsdorf:
Users can now one click anywhere on the project name cell to choose a project and submit.
Project cookie will last 30 days by default.
Project cookie will "remember" by default.

The only question I have is if we want to try to change the project lifetime that users already have set in their config.php. I don't think we have any mechanism to change that file. I suppose we could move it into the setup file though. Thoughts?


glensc:
looks good. let's leave existing config as is.

PR #122, opened 2015-12-02
Add API Tokens support for authentication for RPC/CLI

balsdorf:
This changes the CLI to use an API Key instead of username / password for authentication. This allows it to work with all auth backends (such as CAS).

### DIFF
--- a/lib/eventum/auth/APIAuthToken.php
+++ b/lib/eventum/auth/APIAuthToken.php
@@ -1,0 +1,153 @@
+<?php
+
+/* vim: set expandtab tabstop=4 shiftwidth=4 encoding=utf-8: */
+// +----------------------------------------------------------------------+
+// | Eventum - Issue Tracking System                                      |
+// +----------------------------------------------------------------------+
+// | Copyright (c) 2015 Eventum Team.                                     |
+// |                                                                      |
+// | This program is free software; you can redistribute it and/or modify |
+// | it under the terms of the GNU General Public License as published by |
+// | the Free Software Foundation; either version 2 of the License, or    |
+// | (at your option) any later version.                                  |
+// |                                                                      |
+// | This program is distributed in the hope that it will be useful,      |
+// | but WITHOUT ANY WARRANTY; without even the implied warranty of       |
+// | MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the        |
+// | GNU General Public License for more details.                         |
+// |                                                                      |
+// | You should have received a copy of the GNU General Public License    |
+// | along with this program; if not, write to:                           |
+// |                                                                      |
+// | Free Software Foundation, Inc.                                       |
+// | 51 Franklin Street, Suite 330                                        |
+// | Boston, MA 02110-1301, USA.                                          |
+// +----------------------------------------------------------------------+
+
+use \Firebase\JWT\JWT;
+
+class APIAuthToken
+{
+    private static $default_alg = "HS256";
+
+    public static function generate($usr_id)
+    {
+        $expires = date_add(Date_Helper::getDateTime(), date_interval_create_from_date_string('6 months'));
+
+        $token = JWT::encode(array(
+            'iss'   =>  APP_BASE_URL,
+            'usr_id'    =>  $usr_id,
+            'exp'   =>  $expires->format('U'),
+        ), Auth::privateKey());
+
+        self::saveToken($usr_id, $token, $expires);
+
+        return $token;
+    }
+
+    public static function saveToken($usr_id, $token, $expires)
+    {
+        $sql = "INSERT INTO
+                    {{%api_token}}
+                SET
+                    apt_usr_id = ?,
+                    apt_created = ?,
+                    apt_token = ?,
+                    apt_expires = ?";
+        try {
+            $res = DB_Helper::getInstance()->query($sql, array(
+                $usr_id,
+                Date_Helper::getCurrentDateGMT(),
+                $token,
+                Date_Helper::getSqlDateTime($expires)
+            ));
+        } catch (DbException $e) {
+            return -1;
+        }
+    }
+
+    public static function checkTokenStatus($token)
+    {
+        $sql = "SELECT
+                    apt_usr_id
+                FROM
+                    {{%api_token}}
+                WHERE
+                    apt_token = ? AND
+                    apt_status = 'active' AND
+                    apt_expires > ?";
+        try {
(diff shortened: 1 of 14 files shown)


glensc:
please rebase to avoid merge commit

glensc:
omg, the api keys is really long. does it need to be that long?

glensc:
i would still keep username in the requests. this way can use http authorization header in eventum-rpc client code via [setCredentials()](https://github.com/eventum/rpc/blob/v3.0.3/class.Eventum_RPC.php#L68-L78).

glensc:
also i understand the api keys expire, is that neccessary?

glensc:
eventum secret key - is it neccessary to use it? can't just use [generateRandom](https://github.com/eventum/eventum/blob/v3.0.6/lib/eventum/class.misc.php#L986-L1002) as api keys storage?

that makes whole JWT unneccessary

glensc:
also composer.json seems to have syntax error

lib/eventum/auth/APIAuthToken.php
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        return $usr_id;
+    }
+
+    public static function getUserIDByToken($token)
+    {
+        try {
+            // check that token is valid and not expired
+            $decoded = JWT::decode($token, Auth::privateKey(), array(self::$default_alg));
```
glensc: um, you have still JWT in your code.

lib/eventum/auth/APIAuthToken.php
```diff
@@ -0,0 +1,159 @@
...
+                return self::getTokensForUser($usr_id, false);
+            } else {
+                return array();
+            }
+        }
+
+        return $res;
+    }
+
+    public static function regenerateKey($usr_id)
```
glensc: shouldn't the method include token to invalidate? so one user could have several active tokens, or what's the idea otherwise of having token status?

upgrade/patches/43_api_token.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE {{%api_token}} (
+  apt_id int unsigned NOT NULL auto_increment,
+  apt_usr_id int unsigned NOT NULL,
+  apt_created datetime NOT NULL,
+  apt_status varchar(10) NOT NULL default 'active',
+  apt_token varchar(32) NOT NULL,
+  PRIMARY KEY (apt_id),
+  KEY (apt_token)
```
glensc: i think the KEY should be usr_id+apt_token and should select too from database using such WHERE

glensc: altho probably for security, should probably never include token in the sql query (except unavoidable INSERT), and if comparing for for value select all active tokens for usr_id, they key should be usr_id+apt_status instead.

lib/eventum/auth/APIAuthToken.php
```diff
@@ -0,0 +1,159 @@
...
+            // make sure token hasn't been revoked
+            $usr_id = self::checkTokenStatus($token);
+
+            return $usr_id;
+        } catch (Exception $e) {
+            throw new AuthException($e);
+        }
+    }
+
+    public static function getTokensForUser($usr_id, $auto_generate = false)
```
glensc: i would move the auto_generate logic out from here, and instead add parameter $active = true, i.e whether to fetch only active tokens or all of them.

lib/eventum/auth/APIAuthToken.php
glensc: seems getUserIDByToken should be dropped as it is not in use

lib/eventum/auth/APIAuthToken.php
```diff
@@ -0,0 +1,159 @@
...
+        try {
+            if (self::checkTokenStatus($token) == User::getUserIDByEmail($email, true)) {
+                return true;
+            }
+        } catch (AuthException $e) {
+            return false;
+        }
+    }
+
+    public static function checkTokenStatus($token)
```
glensc: should include $usr_id in functin call, and fetch all user tokens with status active and compare them php side using not time leaking algorithm. but that method is private right now in AuthPassword method. so it can be done later when this PR is merged.

lib/eventum/auth/APIAuthToken.php
balsdorf: Yeah since it was unused I didn't notice that JWT code anymore. It is dropped now (I swapped the names on a few methods)

lib/eventum/auth/APIAuthToken.php
balsdorf: I don't think we need multiple active tokens, at least not until we use tokens for more then just the CLI.

The point of having token status is so we can invalidate old tokens but still have them for reference if needed.

upgrade/patches/43_api_token.sql
balsdorf: Having an auto increment ID makes life easier and pretty much every other table in eventum uses it. 

As far as excluding the token from the SQL query, what is your reason for this? Since we aren't hashing the token it shouldn't be vulnerable to a timing attack.

lib/eventum/auth/APIAuthToken.php
balsdorf: I think this is the best place for the auto generate logic. Otherwise it is moved out into the controller and if we need to auto generate a token in some other place we will have to duplicate it.

No code would use the $active parameter at this point and I prefer not to add dead code.

upgrade/patches/43_api_token.sql
glensc: to exclude token from queries, can avoid making tokens visible from `mysqladmin processlist`

glensc: where do you say to ditch auto increment? i only said `KEY (apt_token)` should be rather `KEY (apt_token, apt_status)`

but as i see you disagree with the rest, probably can leave as is and merge when you're done.

glensc:
merge?

upgrade/patches/43_api_token.sql
balsdorf: Ah, sorry I always think of KEY just as "primary key" and INDEX for other keys.  I'll add it.

Your comment about mysqladmin makes sense for people on shared hosts, etc and I'll add that.

balsdorf:
Made a few tweaks and am ready for it to be merged.

On Tue, Dec 8, 2015 at 10:48 AM, Elan Ruusamäe <notifications@github.com>
wrote:

> merge?
>
> —
> <https://github.com/eventum/eventum/pull/122#issuecomment-162942029>.
>



-- 
Bryan Alsdorf

PR #116, opened 2015-11-28
show dates human friendly

glensc:
uses [jquery-timeago](http://timeago.yarp.com/) to print dates human readable.
to switch between human readable time ago and full timestamp, just click on the element.

could be done, but not (yet?):
- user preference to disable `timeago` formatting
- localized `timeago` messages. localized strings exist in [jquery-timeago](https://github.com/rmm5t/jquery-timeago/tree/v1.4.3/locales) project

how to check?
- code should not use `Date_Helper::getFormattedDate` method
- templates should use `{$var|format_date}` or `{$var|timeago}` modifier

----

before:
before

after:
after2

PR #125, opened 2016-01-08
Added setup option to disable timeago

balsdorf:
Our internal users do not like time ago so I added a setup option to disable it. It is enabled by default.

### DIFF
--- a/htdocs/js/main.js
+++ b/htdocs/js/main.js
@@ -103,7 +103,7 @@ $(document).ready(function() {
         $el.attr('title', $el.text());
         $el.text(old);
     };
-    if ($head.data('config-timeago') == 'enabled') {
+    if ($head.data('config-relative-date') == '1') {
         // if enabled, then enable for all elements
         $timeago
             .timeago()
--- a/htdocs/js/main.js
+++ b/htdocs/js/main.js
@@ -39,7 +39,8 @@ $(document).ready(function() {
 
     ExpandableCell.ready();
 
-    Eventum.rel_url = $('head').attr('data-rel-url');
+    var $head = $('head');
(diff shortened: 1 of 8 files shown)


glensc:
it should came from user prefs instead, like noted in original PR implementing the feature: #116

having it configured in two places makes it confusing, altho i like the default coming from setup, i.e could be useful for new users, so maybe leave your change in, but add per user preference as well?

and timeago should be always there, just default should be human readable vs raw, so having it off you could still obtain timeago by clicking or hovering.

so what you think? as if so, could do user pref later, still keep this PR as is

glensc:
actually, i went ahead and just implemented the disabled state behaviour, so just user preference option missing

balsdorf:
Thanks for implementing the disabled state. I went ahead and made it a user preference.

I also decided to remove "timeago" from the actual setup file name, changed it to "relative_date" so we aren't tied to one specific library name in the future

glensc:
there's seem to be mix of `enabled`, `1` and `0` values. should stay consistent, use everywhere `enabled` or the numbers, i would prefer numbers, as we may have in the  future value `2` for some specific behavior.

and as i look the code, imho it's broken as to dom you write `enabled`, but in `main.js` you check for `1`

balsdorf:
setup uses "enabled" while user preferences uses 1 and 0. I don't like it either but wanted to stay consistent with each section. Changing setup to use 1 and 0 was beyond the scope of this PR.

I write the value of upr_relative_date to the DOM which will be 1 or 0. For existing users I check if "enabled" == 1 but for new users I was just using the string so that was broken. Fixed.

glensc:
you still need to cast it to `int`. while boolean `true` converted to string is `1`, then boolean `false` converted to string is empty string.

PR #127, opened 2016-01-14
do issue duplication and close in same step

glensc:
never liked the way issue duplication is handled, always have to do separate close issue after issue being marked as duplicate. as a result people tend just close issues without bothering duplication status, as setting state as duplication itself does nothing.

this is first step, i guess the close form could be made intelligent to see if resolution is marked as 'duplicate' would ask for duplicate issue id as well


glensc:
I wonder what was missing here for merge, the changes look good here.

@balsdorf

glensc:
while testing, made some changes:
- * 2d70935e8 - use generic error page. "auth_customer" not found in any template (10 seconds ago)

@balsdorf can you check that `$this->role_id == User::ROLE_CUSTOMER` check is superfluous?
- https://github.com/eventum/eventum/blob/v3.5.6/src/Controller/CloseController.php#L73-L74

glensc:
i think `Issue::exists` seems pointless, as access check won't let access this far
- https://github.com/eventum/eventum/blob/v3.5.6/src/Controller/CloseController.php#L96-L100

glensc:
Extra: could prefill status/resolution as killed/duplicate in close form

PR #131, opened 2016-01-26
reuse existing Routing::route to route mails from support::getEmailInfo

glensc:
reuse existing Routing::route to route mails from support::getEmailInfo
some methods made private as they are no longer used elsewhere.

i have not tested the code yet, it's hard to test this in dev environment, but i compared the methods and they seem to do similar thing

there was some confusion with `ema_leave_copy` handling. i think it got in broken in initial commit dc0bf7185 (Eventum 2.2), which is probably now fixed with 137bd2b943a

### DIFF
--- a/lib/eventum/class.routing.php
+++ b/lib/eventum/class.routing.php
@@ -35,15 +35,11 @@ class Routing
 
         // we create addresses array so it can be reused
         $addresses = array();
-        if (isset($structure->to)) {
-            foreach ($structure->to as $address) {
-                $addresses[] = $address->mailbox . '@' . $address->host;
-            }
+        if (isset($structure->headers['to'])) {
+            $addresses[] = $structure->headers['to'];
         }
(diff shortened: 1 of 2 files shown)


glensc:
what is more weird, is that Admin->Manage emails says:

```
[x] Use account for non-subject based email/note/draft routing. 
Note: If you check this, you cannot leave a copy of messages on the server.
```

so checking for `ema_leave_copy` state seems pointless in `ema_use_routing` check... no?

glensc:
tested note routing. seems working, so likely the other routes do too

balsdorf:
Initial look at code seems good and I'll do some testing on this over the next few days.

balsdorf:
I just tested this a bit, it looks good!

> so checking for ema_leave_copy state seems pointless in ema_use_routing check... no?

Yes, I guess this check is pointless.

PR #133, opened 2016-01-28
fix wrapping the long lines

slay123:
Prevent long lines make width very long, wrap them to the next line. Based on CSS provided by Risto Risti.

Inside issue view:
issue

Note view:
note


glensc:
LGTM

Issue #102, opened 2015-11-12
Password for pop account are plain text in the database


plindry:
Like user passwords i consider the email passwords must to be hashed in the database.
Consider this like an enhacement.

cpinfold:
It would be nice to be able to do that.

The POP server requires an unhashed password (being an older standard).  If
you are using Google as an example, you can setup an application specific
password for Eventum though.

Craig

glensc:
Eventum needs plain text to authenticate to POP3/IMAP, as it acts like mail client like your mail agent. You should be using TLS to enhance security.

plindry:
I think you wan't understandme. Yes of course you can use TLS. 

Eventum as you when you talk with the mail server need plain text but not the database and the database administrator needs this kinds of information as plain text saved . Eventum can get the hash, decript and use the plain text with TLS if you want.

glensc:
you're speaking of symmetric encryption, not one way hashing then! passwords are hashed in database, which is one way only! https://en.wikipedia.org/wiki/Cryptographic_hash_function

plindry:
Yes symetric encryption, I Know is not a very secure way but it's better than plain text.
I supposed that the passwords are saved with symetric encription. In my mind i translated hashed passwords as the result of a symetric encryption in two ways. (my fault)

glensc:
no, passwords are hashed

https://github.com/eventum/eventum/releases/tag/v3.0.4
https://github.com/eventum/eventum/pull/77

plindry:
Nice job

glensc:
#134 is now merged

glensc:
btw, do you actually use this feature now?

PR #134, opened 2016-01-28
symmetric encryption

glensc:
this is something that was addressed in #102

image

- [x] class for encrypt/decrypt
- [x] use own key storage or existing private_key? -> will use new file
- [x] modify admin interface not to spit out existing passwords (if no password is provided, use existing one what is stored)
- [x] map places that store passwords
  - db: [email_account.ema_password](https://github.com/eventum/eventum/blob/v3.0.8/lib/eventum/class.email_account.php#L260)
 - config/setup.php: `database.password`
 - config/setup.php: `smtp.auth`
 - config/ldap.php: `bindpw`
- [x] figure out new db field length need ([255 is fine](https://github.com/eventum/eventum/pull/134#issuecomment-180021855))
- [x] re-code all passwords if private key is changed
- [x] make encryption optional?
- [x] figure out migration
 - for config, it's transparent (https://github.com/eventum/eventum/pull/134#issuecomment-176836486)
 - for database we run conversion once (see upgrade/patches/encrypt_passwords.php)
 - the feature will be optional
- [ ] allow enabling encryption from setup page


glensc:
Config keys encryption is actually very cool:

[var_export](http://php.net/var_export) used to serialize `config/setup.php` produces `__set_state` into config file:
```php
    'password' =>                                                                                                        
    Eventum\Crypto\EncryptedValue::__set_state(array(                                                                    
       'ciphertext' => '5rphRKB88tTAWlu5VIRY+K+tneoOtm3EmuW5FuCth9qnjM9ZSJVAsZvbGpqwP5gmjlQKDLinw+x1Z70iKDXUQA',         
    )),      
```

[__set_state](https://github.com/eventum/eventum/commit/6abbf7dec350699d13db03f99df1982e250ade22) method will create new `EncryptedValue` object when deserializing

as a result, any Config key converted to `EncryptedValue` will be transparently handled in further loading/saves.

a [patch](https://github.com/eventum/eventum/commit/35f92c90185fcf87ed29899fbf5895e968e5ae8c) will set initial keys that should be encrypted and they stay that way.

glensc:
```
mysql> select length(ema_password) from eventum_email_account;
+----------------------+
| length(ema_password) |
+----------------------+
|                  192 |
|                   86 |
+----------------------+
2 rows in set (0.00 sec)
```

255 seems enough for encrypted storage, that 86 is typical 12-byte length password encrypted, 192 is 86 byte password :)

glensc:
**Troubleshooting** (for wiki)

In case there's problem, you can edit `config/setup.php` and replace encrypted value with plain text string. To re-crypt again, go to admin panel disable encryption and enable it again. this will re-crypt the values.

plindry:
Nice work.

Thanks

PR #135, opened 2016-01-29
Allow users to be in multiple groups

balsdorf:
This change allows users to be in more then one group at the same time.


glensc:
looks good, altho i don't use the feature due the fact it's restricted to single value.

PR #140, opened 2016-02-09
Add Message-Id column to mail queue table

glensc:
Add Message-Id column to mail queue table

idea is to implement 5d043aaccbee438091081232125684fe3a6e0da5 better, using Message-Id
see discussion in Launchpad
https://bugs.launchpad.net/pld-linux/+bug/1195856

This depends on #139, so it should be merged first

### DIFF
--- a/upgrade/patches/48_maq_message_id.php
+++ b/upgrade/patches/48_maq_message_id.php
@@ -23,11 +23,11 @@ use Zend\Mail\Headers;
 
 $logger = Logger::getInstance('db');
 
-$db->query("alter table {{%mail_queue}} add maq_message_id varchar(255) DEFAULT NULL AFTER maq_subject");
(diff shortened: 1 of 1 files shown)


glensc:
did some query, some types of mails do not include message-id (or headers corrupted and not parseable):

```
mysql> select maq_type,count(*) from q_maq_null_2015 group by 1 order by 2 desc;
+--------------------+----------+
| maq_type           | count(*) |
+--------------------+----------+
| error              |     7453 |
|                    |      300 |
| new_issue          |       44 |
| auto_created_issue |       25 |
| updated            |       24 |
| other_email        |       16 |
| closed             |       10 |
| notes              |        6 |  
| assignment         |        5 |
| blocked_email      |        2 |
+--------------------+----------+
10 rows in set (0.01 sec)
```

the missing message-id will be solved however with #139

PR #143, opened 2016-02-14
collapse replies in email

glensc:
collapse replies and email signatures like gmail does

image

uses [willdurand/email-reply-parser](https://packagist.org/packages/willdurand/email-reply-parser) for that

for expandable cell, even [jQuery.on](http://api.jquery.com/on/) is supposed to work for html that is loaded after, i could not get it to work, so I just setup it again when expandable cell is loaded.

does this has to have user preference again?


glensc:
found one bug: https://github.com/willdurand/EmailReplyParser/pull/42

balsdorf:
Nice feature! I know internally some of our users would like this to be a preference.

glensc:
so, no ideas how to make jQuery `.on` work without hacks? i.e the old `.live()` method was supposed to do that?

balsdorf:
Are you referring to making .on work with elements that are loaded via ajax? I don't see any way to do it easily. What I've done in the past is just had each ajax call re-init the event handling code which is kinda hackish but works with namespaces.

glensc:
Yes, exactly, and the re-init is the current approach too.

glensc:
fixed https://github.com/eventum/eventum/pull/143#issuecomment-184166827
by using our own signature regexp. requires unreleased code, but i added already dev branch in deps for now.

glensc:
Another problem found. diff(1) blocks get rendered as signatures:

```diff
--- eventum/composer.json~    2016-02-15 12:37:40.044380114 +0200
+++ eventum/composer.json    2016-02-29 11:38:07.439787867 +0200
@@ -81,19 +81,17 @@
         "php": ">=5.3.23",
         "php-gettext/php-gettext": "1.0.*",
         "phplot/phplot": "~6.1.0",
+        "ramsey/array_column": "~1.1.3",
```

PR #145, opened 2016-02-16
Autosave notes/emails to local storage in case of browser crash

balsdorf:
Decided to try this approach as it was simple to implement vs using ajax to save client side.

I don't see the need for a user preference on this unless someone else does?

### DIFF
--- a/templates/post_note.tpl.html
+++ b/templates/post_note.tpl.html
@@ -68,9 +68,8 @@
         $('#post_note_form').garlic({
             getPath: function ( $elem ) {
               var form = $elem.parents('form');
-              return form.find('input[name=issue_id]').val() + '_' +
-                      form.find('input[name=parent_id]').val() +
-                      form.data('sup-id') + $elem.attr( 'name' );
+              return [form.find('input[name=issue_id]').val(), form.find('input[name=parent_id]').val(),
+                  form.data('sup-id'),$elem.attr('name')].join('_');
             }
         });
     });
--- a/templates/post_note.tpl.html
+++ b/templates/post_note.tpl.html
@@ -67,7 +67,10 @@
 
         $('#post_note_form').garlic({
             getPath: function ( $elem ) {
-                return $elem.parents('form').find('input[name=issue_id]').val() + '_' + $elem.attr( 'name' );
+              var form = $elem.parents('form');
+              return form.find('input[name=issue_id]').val() + '_' +
+                      form.find('input[name=parent_id]').val() +
+                      form.data('sup-id') + $elem.attr( 'name' );
             }
         });
     });
@@ -75,7 +78,7 @@
 </script>
 
 
-<form class="validate" data-validation-function='validateForm' id="post_note_form" name="post_note_form" method="post">
+<form class="validate" data-validation-function='validateForm' id="post_note_form" name="post_note_form" method="post" data-sup-id="{$sup_id|default:''}">
     <input type="hidden" name="cat" value="post_note">
     <input type="hidden" name="parent_id" value="{$parent_note_id|default:''}">
     <input type="hidden" name="issue_id" value="{$issue_id}">
@@ -192,7 +195,7 @@
         <tr class="buttons">
             <td colspan="2">
                 <input name="main_submit_button" type="submit" value="{t}Post Internal Note{/t}">
-                <input type="button" value="{t}Cancel{/t}" onClick="window.close();">
+                <input type="reset" value="{t}Cancel{/t}" onClick="window.close();">
 
 
                 {if $core.current_user_prefs.email_signature != "" and $core.current_user_prefs.auto_append_note_sig != 'yes'}
--- a/templates/post_note.tpl.html
+++ b/templates/post_note.tpl.html
@@ -64,6 +64,12 @@
     $().ready(function() {
         $('#clear_note_cc').click(clearExtraRecipients);
         $('#add_email_signature').click(setSignature);
+
+        $('#post_note_form').garlic({
+            getPath: function ( $elem ) {
+                return $elem.parents('form').find('input[name=issue_id]').val() + '_' + $elem.attr( 'name' );
+            }
+        });
     });
     //-->
 </script>
@@ -121,6 +127,7 @@
             <td>
                 <select multiple name="note_cc[]"
                         data-placeholder="{t}Choose Recipients...{/t}" class="chosen-select"
+                         data-storage="false"
                         >
                     {html_options options=$users}
                 </select>
@@ -131,8 +138,8 @@
                 {t}Add Extra Recipients To Notification List?{/t}
             </th>
             <td>
-                <label><input type="radio" name="add_extra_recipients" value="yes"> {t}Yes{/t}</label>
-                <label><input type="radio" name="add_extra_recipients" value="no" checked> {t}No{/t}</label>
+                <label><input type="radio" name="add_extra_recipients" value="yes" data-storage="false"> {t}Yes{/t}</label>
+                <label><input type="radio" name="add_extra_recipients" value="no" checked data-storage="false"> {t}No{/t}</label>
             </td>
         </tr>
         <tr>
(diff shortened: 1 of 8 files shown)


glensc:
- I can no longer use reply feature, content is replaced with autosave.
- The cancel button on send dialog should also cancel the autosave data
- What if i have multiple send-note dialogs open, are they all saved differently?

balsdorf:
Good catches. Reply feature is now fixed. A unique save key is generated from the issue ID and reply ID (if it exists) so every combination should be saved individually.
Cancel now clears the auto save data. I've removed the "Reset" button from send email as it was broken (called a missing JS function) and I can't see any need for it.
Multiple send note dialogs will save differently, as long as they are for different issues or reply ids

glensc:
i do not see your changes, did you forgot to push?

balsdorf:
damnit.

Pushed.

On Wed, Feb 17, 2016 at 2:16 PM, Elan Ruusamäe <notifications@github.com>
wrote:

> i do not see your changes, did you forgot to push?
>
> —
> <https://github.com/eventum/eventum/pull/145#issuecomment-185384447>.
>



-- 
Bryan Alsdorf

glensc:
seems working!

altho at first time i tested, i noticed user dropdown being replaced with word "Cancel", i can not see this bug anymore.

templates/post_note.tpl.html
```diff
@@ -64,12 +64,21 @@
...
         $('#clear_note_cc').click(clearExtraRecipients);
         $('#add_email_signature').click(setSignature);
+
+        $('#post_note_form').garlic({
+            getPath: function ( $elem ) {
+              var form = $elem.parents('form');
+              return form.find('input[name=issue_id]').val() + '_' +
+                      form.find('input[name=parent_id]').val() +
+                      form.data('sup-id') + $elem.attr( 'name' );
+            }
```
glensc: i think should always specify the separator between elements to avoid same id for:
- parent_id 1 + sup_id 11
- parent_id 11 + sup_id 1

[ item1, item2 ].join('_') ?

glensc:
so, the library will handle automatically types "submit" and "reset" to do the right thing (clear the autosave data)?

glensc:
when i hit "reply" in issue and close window, it warns me about unsaved, yet it does not warn about unsaved when i hit "reply as note". perhaps remove the warning now that it's autosaved? or change it to warn that "attached files are not autosaved" altho they could be saved as the files are already uploaded to server, form contains attachment id's.

slay123:
Hey.

Tested reply to issue and reply to note. All replies are saved each note has different "save slot", cancel clears the buffer, just as expected.

Can we do the same thing for timetracking. There has been lots of accidents when opening timetracking again and losing correct start time or "Summary".

templates/post_note.tpl.html
balsdorf: That was my plan, but I missed it. using join is much cleaner looking as
well. Done

On Thu, Feb 18, 2016 at 5:18 AM, Elan Ruusamäe <notifications@github.com>
wrote:

> In templates/post_note.tpl.html
> <https://github.com/eventum/eventum/pull/145#discussion_r53301177>:
>
> > @@ -64,12 +64,21 @@
> >      $().ready(function() {
> >          $('#clear_note_cc').click(clearExtraRecipients);
> >          $('#add_email_signature').click(setSignature);
> > +
> > +        $('#post_note_form').garlic({
> > +            getPath: function ( $elem ) {
> > +              var form = $elem.parents('form');
> > +              return form.find('input[name=issue_id]').val() + '_' +
> > +                      form.find('input[name=parent_id]').val() +
> > +                      form.data('sup-id') + $elem.attr( 'name' );
> > +            }
>
> i think should always specify the separator between elements to avoid same
> id for:
>
>    - parent_id 1 + sup_id 11
>    - parent_id 11 + sup_id 1
>
> [ item1, item2 ].join('_') ?
>
> —
> <https://github.com/eventum/eventum/pull/145/files#r53301177>.
>



-- 
Bryan Alsdorf

balsdorf:
@glensc Yes, "submit" and "reset" are handled automatically by the library. I've gone ahead and changed the warning message to be about files. We never had one on notes so I don't think we need to add it.

@slay123 Thanks for testing! Since it is so simple I've gone ahead and added it for time tracking and phone calls (for consistency)

slay123:
Great. Tested timetracking and phone aswell, seems working perfectly :)

slay123:
Found small annoyance. When I add timetracking (click the button), then the save won't get cleared. At least when using bookmarklet.

ie.
javascript:void(window.open('https://eventum./time_tracking.php?iss_id=XXX',%20'tt',%20'width=550,height=250,top=270,left=260,resizable=yes,scrollbars=yes,toolbar=no,location=no,menubar=no,status=no'));

balsdorf:
Hmmm, I can't reproduce that,

However I did find I stupidly did not make time tracking / phone calls
unique per issue and now fixed it. That might fix your issue.

On Fri, Feb 19, 2016 at 6:15 AM, Kristo Klausson <notifications@github.com>
wrote:

> Found small annoyance. When I add timetracking (click the button), then
> the save won't get cleared. At least when using bookmarklet.
>
> ie.
> javascript:void(window.open('
> https://eventum./time_tracking.php?iss_id=XXX',%20'tt',%20'width=550,height=250,top=270,left=260,resizable=yes,scrollbars=yes,toolbar=no,location=no,menubar=no,status=no
> '));
>
> —
> <https://github.com/eventum/eventum/pull/145#issuecomment-186193223>.
>



-- 
Bryan Alsdorf

glensc:
@balsdorf when you commit fix which is related to PR, but PR already merged (or commit not the same branch), include the ticket id as last thing of first line of commit message: ``fixes this and that. #145``

slay123:
@balsdorf - unfortunately it didn't help.

When I open the timetracking window, it will save the timestamp, message etc, so when browser crashes or I accidentially close, refresh or re-open the window, it restores the saved settings, that is very okay.
But when I finished my task and I enter correct time and click "Add Time Entry", then I'm done, I don't need the restoration anymore.
But whenever I re-open the timetracking window, it still restores the last session that I don't really need anymore.

Sample.
* I fill timetracking window and click on the button because I am done with this timetracking:
screenshot 2016-02-25 12 50 28

* Time will get inserted into issue:
screenshot 2016-02-25 12 50 45

* When I need NEW timetracking and click Add Time Entry I will get window with old and expired data instead of EMPTY timetracking window:
screenshot 2016-02-25 12 50 52

glensc:
Re: kristo's bug

[templates/add_time_tracking.tpl.html](https://github.com/eventum/eventum/blob/0d519b9aeccab05feaaa94ac1c0b2e78f4db5ae3/templates/add_time_tracking.tpl.html#L133) already has `type=submit`, could be that [submit handler](https://github.com/eventum/eventum/blob/0d519b9aeccab05feaaa94ac1c0b2e78f4db5ae3/templates/add_time_tracking.tpl.html#L76) breaks garlic?

balsdorf:
Yup, had just tested this. For me, applying the garlic handler first solves
the problem.

On Thu, Feb 25, 2016 at 2:10 PM, Elan Ruusamäe <notifications@github.com>
wrote:

> Re: kristo's bug
>
> templates/add_time_tracking.tpl.html
> <https://github.com/eventum/eventum/blob/0d519b9aeccab05feaaa94ac1c0b2e78f4db5ae3/templates/add_time_tracking.tpl.html#L133>
> already has type=submit, could be that submit handler
> <https://github.com/eventum/eventum/blob/0d519b9aeccab05feaaa94ac1c0b2e78f4db5ae3/templates/add_time_tracking.tpl.html#L76>
> breaks garlic?
>
> —
> <https://github.com/eventum/eventum/pull/145#issuecomment-188959073>.
>



-- 
Bryan Alsdorf

glensc:
heh, i missed you with 2 minutes, my "fix" was to use html based handler :)

glensc:
found a way to reproduce weirdness i got when i initially tested this

  - click send note button
  - type something to Extra Note Recipients (something that does not actually complete)
  - close the window
  - reopen the window by pressing same send note button
  - whatever you typed there appears as caption of "cancel" button

PR #146, opened 2016-02-17
Logger config

glensc:
Adds support for configuring loggers via config file using [Monolog-Cascade](https://github.com/theorchard/monolog-cascade)

Main goal was to be able to add more loggers without modifying code, for example log to Slack, etc

Current approach loads:
 - config/logger.dist.php
 - config/logger.php, this file can be missing

now how to ensure people don't go and modify `config/logger.dist.php` and then get their changes lost on upgrade? or should the `config/logger.dist.php` file be moved out of config dir, to `res/` maybe?

i've chosen .php format, as it allows more flexible customization


glensc:
for example i have in my local `logger.php`:

```php

if ($setup['email_error']['status'] == 'enabled') {
    // enable error mailer
    $config['handlers']['error_mailer'] = array(
        'class' => 'Eventum\\Monolog\\MailHandler',
        'level' => 'ERROR',
    );
    $config['loggers']['app']['handlers'][] = 'error_mailer';
}
```

how to share such snippets? or have these automatic?

balsdorf:
This is great.

As far as people modifying logger.dist.php, yeah I'd move it out of config/ as people are used to changing things there.  I'd also add a huge, ugly warning to not change it.

For sharing snippets, we could have docs/examples/config/logger.php with all of them or a logger/ subdir.

glensc:
so, `res/logger.php` ?

balsdorf:
Yes

On Fri, Feb 19, 2016 at 2:18 AM, Elan Ruusamäe <notifications@github.com>
wrote:

> so, res/logger.php ?
>
> —
> <https://github.com/eventum/eventum/pull/146#issuecomment-186113940>.
>



-- 
Bryan Alsdorf

glensc:
moved, but i think `$setup['email_error']['status']` should be somehow automatically setup, not sure how

PR #149, opened 2016-03-02
Allow separate role for editing custom fields vs viewing.

balsdorf:


balsdorf:
I need to do some additional testing but wanted to get this out there.

glensc:
looks simple. but let us know when you feel it's ready, or merge yourself

glensc:
please test what happens if someone sets edit_role < view_role and perhaps also show error in ui when someone does?

balsdorf:
It actually did not cause any problems but I've added validation.

I will do some additional testing and merge tomorrow most likely.

PR #150, opened 2016-03-04
Don't clear saved data until after the next page actually loads,

balsdorf:
#145 enchancement:

This prevents network / server errors from causing saved data to be lost.

Should be good to merge, but wanted others to test it first

### DIFF
--- a/src/Controller/PostNoteController.php
+++ b/src/Controller/PostNoteController.php
@@ -100,7 +100,7 @@ class PostNoteController extends BaseController
 
         if ($this->cat == 'post_result' && ($post_result = $get->getInt('post_result'))) {
             $this->tpl->assign('post_result', $post_result);
-            $this->tpl->assign('garlic_prefix', $get->has('garlic_prefix') ? $get->get('garlic_prefix') : '');
+            $this->tpl->assign('garlic_prefix', $get->get('garlic_prefix'), '');
         } elseif ($this->cat == 'post_note') {
             $this->postNoteAction();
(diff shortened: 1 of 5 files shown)


src/Controller/PostNoteController.php
```diff
@@ -100,6 +100,7 @@ protected function defaultAction()
 
         if ($this->cat == 'post_result' && ($post_result = $get->getInt('post_result'))) {
             $this->tpl->assign('post_result', $post_result);
+            $this->tpl->assign('garlic_prefix', $get->has('garlic_prefix') ? $get->get('garlic_prefix') : '');
```
glensc: if you don't have "0" there as valid value, you can write this as:

`$get->get('garlic_prefix', '')` to have default value empty string

templates/post_note.tpl.html
```diff
@@ -65,22 +68,28 @@
         $('#clear_note_cc').click(clearExtraRecipients);
         $('#add_email_signature').click(setSignature);
 
-        $('#post_note_form').garlic({
-            getPath: function ( $elem ) {
-              var form = $elem.parents('form');
-              return [form.find('input[name=issue_id]').val(), form.find('input[name=parent_id]').val(),
-                  form.data('sup-id'),$elem.attr('name')].join('_');
-            }
+        var post_note_form = $('#post_note_form');
```
glensc: i like the convention that `jQuery` objects are prefixed with `$`

balsdorf:
Updated based on @glensc 's feedback and fixed bug where "cancel" didn't clear data.

Issue #155, opened 2016-03-23
Zend Mail fails on emails with "From XXX" header


balsdorf:
If an email starts with a header such as "From test@example.com  Wed Mar 23 01:42:50 2016" then this exception will be thrown:

`Zend\Mail\Header\Exception\InvalidArgumentException: Invalid header name detected in /media/data/work/eventum-skysql/vendor/zendframework/zend-mail/src/Header/GenericHeader.php on line 57
`

I've attached an example email.

I can't find any reference to this in Zend Mail. 

My two solution ideas are to either filter this line out in the routing class or in MailMessage::createFromHeaderBody(). However I'm not sure if it would also impact us in other areas.
[from_nocolon.txt](https://github.com/eventum/eventum/files/186826/from_nocolon.txt)

glensc:
From where such email comes exactly? downloaded over IMAP? i.e in what codepath in eventum it comes in

as the line you're referring to is in [MBOX](https://en.wikipedia.org/wiki/Mbox) file format definition of "here begins new email"

balsdorf:
This came in through process_all_email.php from postfix.

On Wed, Mar 23, 2016 at 4:42 PM, Elan Ruusamäe <notifications@github.com>
wrote:

> From where such email comes exactly? downloaded over IMAP? i.e in what
> codepath in eventum it comes in
>
> as the line you're referring to is in MBOX
> <https://en.wikipedia.org/wiki/Mbox> file format definition of "here
> begins new email"
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/eventum/eventum/issues/155#issuecomment-200554696>
>



-- 
Bryan Alsdorf
MariaDB Corporation | http://mariadb.com

glensc:
sorry, but no trace possible? currently ZF is used only when composing mails

as i see this is very special case when feeding raw mailbox, the `From ` stripping belongs to routing, everywhere else should be RFC822 compatible mails. but then again, need to see some trace how this kicked in, as if i read the code then i do not see how `$full_email` reaches to ZF class...

glensc:
heh, [Mime_Helper::decode](https://github.com/eventum/eventum/blob/v3.0.10/lib/eventum/class.routing.php#L34) decodes this email as header named `from test@example.com  wed mar 23 01` with value `42:50 2016`:

```php
stdClass::__set_state(array(
   'headers' => array (
[...]

PR #156, opened 2016-03-24
strip out mbox From  header

glensc:
proposed fix for #155


balsdorf:
I've tested this and it works for me. Thanks!

PR #166, opened 2016-04-30
drop table prefix

glensc:
it was suggested by @balsdorf that we drop table prefix.
it was initially added to support on multi hosting.

we believe this is no longer the case, and dropping such support simplifies
code maintenance and development.


glensc:
```
* Applying patch: 57 (57_drop_table_prefix.php)
SQLSTATE[42S02]: Base table or view not found: 1146 Table 'eventum.eventum_version' doesn't exist
```

looks like the upgrade script does the all rename, but fails to mark itself to be "applied":

```
09:40:05 mysql{3}> select * from version where ver_version=57;
Empty set (0.00 sec)
```

however, as the patch is re-entrant, it's harmless (it will just do nothing on retry)

```
# /usr/share/eventum/bin/upgrade.php 
* Applying patch: 57 (57_drop_table_prefix.php)
* Your database is now up-to-date. Version 57
# /usr/share/eventum/bin/upgrade.php 
* Your database is already up-to-date. Version 57
# 
```

glensc:
this should be still fixed as it will stop applying patches, i.e when there are more patches after this one.

PR #168, opened 2016-05-20
WIP: use Symfony Session for session usage

glensc:
switch to use [Symfony Session](http://symfony.com/doc/current/components/http_foundation/sessions.html) classes.

we already have [HttpFoundation](http://symfony.com/doc/current/components/http_foundation/introduction.html) as dependency. and the symfony classes are more up to date with modern php than our own [Session](https://github.com/eventum/eventum/blob/v3.1.0/lib/eventum/class.session.php) class, like supporting PHP 5.4 new [SessionHandlerInterface](http://php.net/manual/en/class.sessionhandlerinterface.php) and [session_register_shutdown](http://php.net/manual/en/function.session-register-shutdown.php)

also solves https://github.com/eventum/eventum/pull/168#issuecomment-220637773


glensc:
`Session::verify($usr_id);` was pointless, it no longer did what it's description said it does. so replaced it's calle with `Session::init($usr_id);`

src/Session.php
```diff
@@ -0,0 +1,74 @@
...
+     */
+    public static function init($usr_id)
+    {
+        $session = static::getInstance();
+
+        // set the IP in the session so we can check it later
+        $session->set('login_ip', $_SERVER['REMOTE_ADDR']);
+
+        // store user ID in session
+        // XXX: Should we perform checks on this usr ID before accepting it?
```
balsdorf: I think we can get rid of this old XXX. Only our trusted code is calling Session::init(). If they can call Session::init() they could just manually set $_SESSION['usr_id'] themselves.

balsdorf:
This looks good to me.

balsdorf:
The problem I had was CAS Authentication stores info in the session BEFORE the eventum user is logged in (and Session::init() is called).

So this is what happened:
* CAS sets $_SESSION data
* Eventum sets Auth Cookie and calls Session::init() which clears all $_SESSION
* Next page load CAS has to re-authenticate (seamless to user but involves another round trip to the CAS server)
* AuthCookie is already set so Session::init() is not called and $_SESSION is not destroyed

The practical problem with this was on the first login to eventum, ajax requests (such as expanding emails) would not work.

So all my old Session::init / verify code was created to solve a theoretical security problem called "session fixation" which really isn't a problem. If you try to research it you see it was a hot topic 5+ years ago but went away as everyone uses SSL and cookies instead of passing session ids in urls.

m39yur1p7:
I just upgraded to 3.1.1 and to make it work had to comment out line 70 in src/Controller/SelectProjectController.php to avoid PHP starting the session before Symfony

glensc:
@m39yur1p7 and what was the error of session started before symfony?

glensc:
@balsdorf can you comment why we have the `session_start` there? i guess the second file is just copy paste of the first one.

  - [src/Controller/SelectProjectController.php:69-70](https://github.com/eventum/eventum/blob/v3.1.1/src/Controller/SelectProjectController.php#L69-L70)
  - [src/Controller/SelectCustomerController.php:58](https://github.com/eventum/eventum/blob/v3.1.1/src/Controller/SelectCustomerController.php#L58)

and whether it's still needed to start session there, if needed i'll add symfony session init there not native php method directly.

m39yur1p7:
@glensc error was: [Fri Jun 03 12:38:59.980379 2016] [:error] [pid 2062] [client ***:61008] app.ERROR: Uncaught Exception RuntimeException: "Failed to start the session: already started by PHP." at /var/https/eventum.***.hu/vendor/symfony/http-foundation/Session/Storage/NativeSessionStorage.php line 135 {"exception":"[object] (RuntimeException(code: 0): Failed to start the session: already started by PHP. at /var/https/eventum.***.hu/vendor/symfony/http-foundation/Session/Storage/NativeSessionStorage.php:135)"} {"url":"/select_project.php","ip":"84.1.57.117","http_method":"GET","server":"eventum.***.hu","referrer":"https://eventum.***.hu/","file":null,"line":null,"class":null,"function":null,"memory_usage":"1.5 MB","memory_peak_usage":"1.5 MB","version":"3.1.1","usr_id":"3"}, referer: https://eventum.***.hu/

glensc:
will be fixed in 3.1.2

balsdorf:
@glensc Just old code that never got cleaned up. 

Since we start sessions automatically now no need for that.

glensc:
yea, i figured it may be needed for some custom customer hooks, but wasn't sure, so i wanted to release a fix that solved fatal error for most common cases. and hoped if you find my question you can tell that it's still needed and can fix it in next version.

PR #170, opened 2016-05-21
Automatically set page ID based on template path and name

balsdorf:
This removes the need to manually set page ids in each template.  This also removes the page id from page classes to remove duplication.

Most page ids stayed the same, but a few have changed to match the template name.


glensc:
LGTM

glensc:
damn, [toggl-button](https://github.com/eventum/toggl-button) got broken due selector change.

PR #177, opened 2016-06-26
Added visible error instead of 500 when workflow class is missing

cpinfold:
Very small change to test whether the workflow class file is found before trying to require it. Now on error shows the standard configuration error instead of a 500 server error.


glensc:
I vote no:
- such internal errors should not be displayed standard user (maybe only administrator)
- you assume requests are text/html that this code path is executed, but there may be requests for image or json responses as well

proper would be to throw exception and each code path could decide what to do with the exception (render as html render as json or png)

glensc:
also doing "exit" in library class is very wrong, makes unit testing impossible.

i would refactor this code path to:
- find the workflow class file, save it to a variable
- at the end of all checks if workflow class file variable is empty throw some exception

i believe exceptions are logged better than fatal errors and thus not solve that part at all in this issue.

PR #190, opened 2016-08-30
Use "in-reply-to" mail header to find Issue ID

cpinfold:
Allowed replies to use non-subject based "in-reply-to" for message-id matching (when subject based routing is enabled).


cpinfold:
@balsdorf @glensc This change has been working very well and I haven't come across any issue, can you merge it?

It completely solves the issue of duplicate issues being created when using subject based routing (where the subject does not contain the issue id).

glensc:
the change does not affect systems which have subject_based_routing=false?

glensc:
also can you please add changelog entry in this PR. we're trying the [merge=union](0dead87443cf3ab1790dd2dfd27a2f314a628324) thing

cpinfold:
@glensc The change adds to the if so that when subject based routing is enabled it still processes the else block if there is no [#IssueID] in the subject line.  
```
and (preg_match("/\[#(\d+)\]( Note| BLOCKED)*/", $subject, $matches)))
```
This allows those who enable subject based routing to also follow the "in-reply-to" message id back to the original issue.  This code was already in place if subject based routing was not enabled.

Changelog also updated.

cpinfold:
Don't merge yet.. trying to clean up the PR.

PR #205, opened 2016-10-05
Reply as email

balsdorf:
Added the ability to reply to internal notes as emails

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,6 +1,7 @@
 # Eventum Issue Tracking System
 
 ## 2016-??-??, Version [3.1.4]
+- Add 'reply as email' option to notes (@balsdorf, #205)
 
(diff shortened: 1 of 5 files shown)


glensc:
please no `else` keywords after `return` statements

also, please also add CHANGELOG entry, let's see how well github handles [merge=union](https://github.com/eventum/eventum/commit/0dead87443cf3ab1790dd2dfd27a2f314a628324)

glensc:
would you mind swapping the columns, so that buttons are ordered as:

`reply as email` `reply as note`

i keep hitting `reply as email` button as it's the usual place to find the button

balsdorf:
Are you talking about the small icons on the view issue page where you have the list of notes or the actual buttons on the view note popup?

glensc:
yes

Issue #202, opened 2016-09-29
process_mail_queue.php error


johnBim:
Hi, 
I try to procress the maill queue and I have the error message bellow. I even tried with the quesry string fix-lock=1, but still the same issue. Could you please help? 
Many thanks

#!/usr/bin/php ERROR: Another instance of the script is still running for the specified account. 
If this is not accurate, you may fix it by running this script with 'fix-lock=1'
in the query string or you may unlock ALL accounts by running this script with 'fix-lock=1'
as the only parameter.

cpinfold:
Hi John,

Do you have access to the command line?  

Are you able to run:
```
/path/to/eventum/bin/process_mail_queue.php --fix-lock
```
(replacing /path/to/eventum with your path)

What is the output of running that command?  Or if through the web server, what does it show?  You should be seeing: "Removed lock file"
How are you causing the process_mail_queue.php to run?  On what frequency?

Craig.

glensc:
running cron scripts from web has been dropped long time ago

you probably should remove the lock file from `var/` dir and retry whatever you are doing. 

i mean you didn't even report what version you are using or how exactly are you invoking the script!

johnBim:
Hi,
basically I have windows hosting and I use Windows Tasks Scheduler to call process_mail_queue.php with browser
I use the version 3.1.2
Thanks

glensc:
did you remove the pid files [as told](https://github.com/eventum/eventum/issues/202#issuecomment-250758504)? if not why so?

glensc:
Remove the lock files and retry. New Eventum version will remove support for lock files when #209 gets merged.

PR #209, opened 2016-10-09
rewrite locking code to use flock

glensc:
i've planning to do this long time ago. the locking can be made really accurate in scope of one machine by using [flock](http://php.net/flock).

this should also solve user errors, who are unable to remove locks: #202

this removes writing pid files and `--fix-lock` arguments as it is no longer necessary.

the current code uses [malkusch/lock](https://github.com/malkusch/lock):
```php
$mutex->synchronized(function () use ($bankAccount, $amount) {
    $balance = $bankAccount->getBalance();
    $balance -= $amount;
    if ($balance < 0) {
        throw new \DomainException("You have no credit.");

    }
    $bankAccount->setBalance($balance);
});
```
the lock is automatically released when the code block exits.

the downside is that the locks are waited forever until the locked process still runs (lock is removed if process is terminated), this may queue up processes if one process stays running for too long.

i bet the error rate of users having stale pid files is way greater than actual situation when one process runs for very long.

however, maybe if lock is finally acquired and it had waited it over 300 seconds, assume lock acquire was unsuccessful and abort?

### DIFF
--- a/bin/irc-bot.php
+++ b/bin/irc-bot.php
@@ -15,13 +15,6 @@ ini_set('memory_limit', '1024M');
 
 require_once __DIR__ . '/../init.php';
 
-// if requested, clear the lock
-if (in_array('--fix-lock', $argv)) {
-    Lock::release('irc_bot');
-    echo "The lock file was removed successfully.\n";
-    exit;
-}
-
 if (in_array('--check-process', $argv)) {
     $check = true;
 } else {
@@ -40,13 +33,4 @@ try {
     exit(1);
 }
 
-// acquire a lock to prevent multiple scripts from
-// running at the same time
-if (!$bot->lock($check)) {
-    error_log('Error: Another instance of the script is still running.');
-    error_log("If this is not accurate, you may fix it by running this script with '--fix-lock' as the only parameter.");
-    exit(1);
-}
-
 $bot->run();
-$bot->unlock();
--- a/bin/irc-bot.php
+++ b/bin/irc-bot.php
@@ -15,12 +15,6 @@ ini_set('memory_limit', '1024M');
 
 require_once __DIR__ . '/../init.php';
 
-if (in_array('--check-process', $argv)) {
-    $check = true;
-} else {
-    $check = false;
-}
-
 // NB: must require this in global context
 // otherise $SMARTIRC_nreplycodes from defines.php is not initialized
 require_once 'Net/SmartIRC/defines.php';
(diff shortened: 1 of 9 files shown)


glensc:
few things came up while developing this:

- [ProcessMailCommand](https://github.com/eventum/eventum/blob/v3.1.3/src/Command/ProcessMailCommand.php) ([process_all_emails.php](https://github.com/eventum/eventum/blob/v3.1.3/bin/process_all_emails.php)) does not have any locking code
- [Monitor::checkIRCBot()](https://github.com/eventum/eventum/blob/v3.1.3/lib/eventum/class.monitor.php#L211-L226) checks for irc bot pid

the new locking code does not write any pids, so the irc bot check will not work.

balsdorf:
> the downside is that the locks are waited forever until the locked process still runs (lock is removed if process is terminated), this may queue up processes if one process stays running for too long.

> i bet the error rate of users having stale pid files is way greater than actual situation when one process runs for very long.

> however, maybe if lock is finally acquired and it had waited it over 300 seconds, assume lock acquire was unsuccessful and abort?

I agree with your thinking, this should be a good solution.

glensc:
i file RFE to the library: https://github.com/malkusch/lock/issues/6

i guess can merge as is for now

glensc:
i already merged this, but those issues did not get addressed:

https://github.com/eventum/eventum/pull/209#issuecomment-252494774

@balsdorf your comments?

glensc:
@balsdorf ping!

balsdorf:
Sorry missed the previous ping.

> - [ProcessMailCommand](https://github.com/eventum/eventum/blob/v3.1.3/src/Command/ProcessMailCommand.php) ([process_all_emails.php](https://github.com/eventum/eventum/blob/v3.1.3/bin/process_all_emails.php)) does not have any locking code

I don't think it needs it as multiple instances of it can run without a problem.

> - [Monitor::checkIRCBot()](https://github.com/eventum/eventum/blob/v3.1.3/lib/eventum/class.monitor.php#L211-L226) checks for irc bot pid

Not sure about this. I personally don't use the monitor class so I'd be okay with dropping it all (or at least this method for the time being). The other option be to write the pid to var/irc_bot.pid or something like that.

glensc:
perhaps i just drop the irc bot pid checking. i use irc bot and that pid check is not valuable. it has even been wrong and creating bogus alerts in the past.

PR #211, opened 2016-10-14
Remove duplicate variables from templates

balsdorf:
All variables set by default are now part of the $core variable.

This might impact 3rd party templates, but the fix is a simple search and replace.


glensc:
I assume you tested this, so merging

PR #212, opened 2016-10-15
Use "in-reply-to" mail header to find Issue ID

cpinfold:
This is a redo of #190 to avoid the merge mess.

Allowed replies to use non-subject based "in-reply-to" for message-id matching (when subject based routing is enabled).

It does this by allowing the existing else clause and references behaviour for non-subject based routing to also occur when the subject line does not contain [#IssueID].

@glensc The change adds to the if so that when subject based routing is enabled it still processes the else block if there is no [#IssueID] in the subject line.  
```
and (preg_match("/\[#(\d+)\]( Note| BLOCKED)*/", $subject, $matches)))
```
This allows those who enable subject based routing to also follow the "in-reply-to" message id back to the original issue.  This code was already in place if subject based routing was not enabled.

Changelog also updated.

I'll close #190 which didn't rebase correctly (my error).

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -11,6 +11,7 @@
 - Throw Exception in CLI if Eventum is not configured (@glensc, 9f04950)
 - Fix Time Tracking administration bugs (@glensc, @yangmx, #197, #196)
 - Add back Authorized Repliers user picker (@glensc, #210)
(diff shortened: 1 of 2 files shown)


glensc:
@cpinfold well. i [asked](https://github.com/eventum/eventum/pull/190#issuecomment-252353716) because i couldn't bother thinking myself does it affect other systems. i'll assume it does not, so merged.

PR #216, opened 2016-10-31
fix SCM checkins being displayed as "public", fixes #215

### LINKED ISSUE #215: SCM Integration section bugs - it's colored blue (Public) and show/hide gives fatal error
1. it is painted blue, as Public section, but de facto is visible only for internal users, see issue #214 for screenshot.
2. clicking show/hide gives JS error

screenshot 2016-10-31 14 38 10

glensc:
fixes #215,
- [x] fix showing block for reporters
- [x] fix js error

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,6 +1,7 @@
 # Eventum Issue Tracking System
 
 ## 2016-??-??, Version [3.1.5]
+- Fix SCM checkins being displayed as "public" (@glensc, #215, #216)
 
 ## 2016-10-26, Version [3.1.4]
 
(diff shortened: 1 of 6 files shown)


glensc:
@balsdorf there's two more places for `toggleVisibility`:
- [templates/impact_analysis.tpl.html:44](https://github.com/eventum/eventum/blob/v3.1.4/templates/impact_analysis.tpl.html#L44)
- [docs/examples/templates/customer/customer_info.tpl.html:13](https://github.com/eventum/eventum/blob/v3.1.4/docs/examples/templates/customer/customer_info.tpl.html#L13)

you can fix them yourself out of scope of this PR, i don't use either of those features.

PR #217, opened 2016-10-31
make public attachments visible for Reporters fixes #214

### LINKED ISSUE #214: Editors cannot see "Attached Files" section and it's also missing Public/Internal switch
Reporter doesn't see any attached files in the issue. This is everything he sees:
screen shot 2016-10-31 at 12 36 31

This is developer view (after issue description):
screenshot 2016-10-31 12 38 09

Attached Files section has lost Internal/Public support and every file is treated as internal item, but still is painted blue. Not sure, if this is desired change but from our POV it's not.

glensc:
make public attachments visible for Reporters, fixes #214

the [Access::canViewAttachedFiles()](https://github.com/eventum/eventum/blame/v3.1.4/lib/eventum/class.access.php#L117-L131) was added in 3836501 and it checked for `Customer` Role. That change was released first as 2.4.0pre1

the change also added `{if $issue_access.files}` condition in [view.tpl](https://github.com/eventum/eventum/commit/38365011f33a19dae2d2816f9fe2b04af882f008#diff-0b7308e0fb8a837778d273399e99a1ee) which was not present before, i.e the `{include file="attachments.tpl.html"}` was loaded unconditionally

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -2,6 +2,7 @@
 
 ## 2016-??-??, Version [3.1.5]
 - Fix SCM checkins being displayed as "public" (@glensc, #215, #216)
+- Make public attachments visible for Viewers (@glensc, #214, #217)
 
 ## 2016-10-26, Version [3.1.4]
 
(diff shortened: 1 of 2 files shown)


glensc:
Role Viewer:
image
Role Admin:
image

PR #223, opened 2016-11-30
Workflow change project

balsdorf:
This can be used to map values for fields between projects or do other custom changes.

### DIFF
--- a/lib/eventum/class.workflow.php
+++ b/lib/eventum/class.workflow.php
@@ -932,11 +932,10 @@ class Workflow
     /**
      * Called when an issue is moved from this project to another.
      *
-     * @param $prj_id
-     * @param $issue_id
-     * @param $new_prj_id
-     * @since 3.1.6
-     * @return null
+     * @param $prj_id integer
+     * @param $issue_id integer
+     * @param $new_prj_id integer
+     * @since 3.1.7
      */
     public static function handleIssueMovedFromProject($prj_id, $issue_id, $new_prj_id)
     {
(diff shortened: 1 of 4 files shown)


glensc:
> Should this be called on the workflow of the old project or the new project?

if you're not sure how, you could add two methods:
- movedFromProject
- movedToProject

so you can choose which one to implement.

balsdorf:
Good idea. I've done that now. I'll do a bit more testing before merge.

Since ChangeLog is conflicting now, should I rebase once everything is good to go?

glensc:
i'm planning to make release shortly, so it will conflict even more. and this PR should be merged after the release, to give it more testing ground.

so yes, rebase after release made (and fix `@since` tags)

lib/eventum/class.workflow.php
```diff
@@ -930,19 +930,37 @@ public static function cryptoDowngradeConfig($prj_id = 1)
...
+    }
+
+    /**
+     * Called when an issue is moved to this project from another.
+     *
+     * @param $prj_id
+     * @param $issue_id
+     * @param $old_prj_id
      * @since 3.1.6
+     * @return null
```
glensc: it doesn't `return` anything. remove this line

lib/eventum/workflow/class.abstract_workflow_backend.php
```diff
@@ -685,22 +685,34 @@ public function cryptoDowngradeConfig()
...
+    }
+
+    /**
+     * Called when an issue is moved to this project from another.
+     *
+     * @param $prj_id
+     * @param $issue_id
+     * @param $old_prj_id
+     * @since 3.1.6
+     */
```
glensc: as minimum, should fill param types

glensc:
relased, you may rebase now

balsdorf:
Well I royally made a mess with git but I think I managed to fix it all.

glensc:
you can see the Commits tab (https://github.com/eventum/eventum/pull/223/commits) what are the actual commits associated, looks good right now.

PR #226, opened 2016-12-09
Don't QP encode sender when storing in the DB. Fixes #225

### LINKED ISSUE #225: From field is garbled if senders name contains unicode
When sending an email from the web interface, if the senders name contains unicode and "Add Unknown Recipients to Issue Notification List" is **not** checked the from field will contain the Quoted-printable version of the name, such as =?UTF-8?Q?Admi=C3=B1_User?= <issue-60@example.com>

image

This is caused by a line of code that is only executed when "Add Unknown" is not checked: https://github.com/eventum/eventum/blob/2415165ab30ab24cd6bed16feb11085705b77720/lib/eventum/class.support.php#L2110

balsdorf:
This stops the sender name from being QP encoded before being stored in the database.  The alternative fix is to always store the QP encoded value and decode it wherever it is displayed.

Since sup_from is for display purposes, I believe storing it in unicode is the correct approach.


glensc:
thanks for spotting this, i had a hunch this is still broken!

and `42_fixencoding.php` needs to be re-ran to fix the db. i wonder should we rename the patch so it gets executed again on upgrades?

balsdorf:
Done.

glensc:
please add the ChangeLog entry

PR #235, opened 2017-01-12
phinx integration

glensc:
https://phinx.org/ - PHP Database Migrations For Everyone

seemed good idea to use something that is not own cooked.

probably need to support both migration tools for some time. invoke both of them from `bin/upgrade.php` and add new migrations to `sphinx` subsystem only.

as it uses PDO classes internally, should probably first drop PEAR db support, and make 3.2 series eventum update to indicate new dependency requirements.

adds new dependencies:
```
  - Installing symfony/debug (v2.8.15) Loading from cache
  - Installing symfony/console (v2.8.15) Loading from cache
  - Installing robmorgan/phinx (v0.6.5) Loading from cache
```

set 3.2.0 milestone where:
- altho [phinx supports table prefix](http://docs.phinx.org/en/latest/configuration.html#table-prefix-and-suffix) we decided to drop table prefix support
- drop pear-db support: #252
- drop old db upgrade mechanism. start from scratch using only phinx engine.


glensc:
@balsdorf @poing @cpinfold which db driver and charset are you using?

cpinfold:
I'm using the PDO driver (manually changed the config) and utf8 charset.

Craig.

glensc:
some issues that have come up when creating initial schema with phinx

- [ ] can not specify length of multikey each elements: https://github.com/robmorgan/phinx/issues/1080. perhaps modify the columns just to be shorter?
- [ ] some tables did not specify `ENGINE` in old patch system, so they would be created with default engine. should we patch existing tables?
 - `upgrade/patches/02_usr_alias.sql`
 - `upgrade/patches/45_multiple_groups.php`
 - `upgrade/patches/07_user_preference.php`

balsdorf:
We can remove the length from the unique index, I'm not sure why it is setup that way and it serves no purpose. It would actually cause a problem if there was a very long hostname.

If we are going to specify an engine, I'd like to specify InnoDB. It even supports fulltext indexes these days. Otherwise, I'd just leave it off and use the default (which on modern installations will be InnoDB). 

Set is the wrong type as it allows multiple values. Either enum or varchar is better.

glensc:
the question about engine was what to do with existing installations.

for new phinx based installation (3.2.x) it's clear that same engine be specified for all tables. (it can be made easily configurable with phinx).

glensc:
as for length in index: no, it must be specified as the sum of field lenghts is too long for index. my suggestion was to truncate fields themselves to be shorter so that length does not need to specified for `KEY`.

so the real question is, what lengts to use that sum fits under <1000 bytes.

i point out the error itself (reported into https://github.com/robmorgan/phinx/issues/1080):
```
SQLSTATE[42000]:
Syntax error or access violation:
1071 Specified key was too long; max key length is 1000 bytes  
```

glensc:
@balsdorf if you think we should change the `SET` field type, could you please give `ALTER` statement how to change it

```mysql
  `prj_status` set('active','archived') NOT NULL DEFAULT 'active',
```

glensc:
probably makes sense to drop `APP_CHARSET` support as well for 3.2.0, i.e support only utf8 (or even utf8mb4)

glensc:
osc (online schema change) topic
- https://github.com/robmorgan/phinx/issues/542
- https://github.com/masom/lhm_php

if ever doing `ALTER` on table with lots of data.

glensc:
unable to change primary key and keeping `auto_increment` to anything else than `signed int`. we have `unsigned` columns, and somewhere even `tinyint`

- [ ] https://github.com/robmorgan/phinx/issues/28

glensc:
those patches that lacked ENGINE. decided to add patch because they lacked also charset definition which means it was likely default as `latin1`. 

- upgrade/patches/68_charset.sql

glensc:
still @balsdorf waiting for your feedback. the branch is almost ready so you can as well test yourself if you wish.

balsdorf:
This is the alter table I'd use for prj_status:

`ALTER TABLE project CHANGE COLUMN prj_status prj_status varchar(8) NOT NULL default 'active';`

I'm not sure how to handle engine for existing installations. InnoDB is a much superior engine to MyISAM but until recent versions did not support fulltext indexes. If someone is using an older version (5.5 isn't that old) and fulltext indexes the conversion to InnoDB would fail. Additionally, if they are used to everything being MyISAM and backing up by locking tables and copying data files converting to InnoDB could screw them.

If possible, I'd make this an optional step to convert to InnoDB, otherwise just leave as is.

I'll check out the branch and see what I think

balsdorf:
I fixed a very minor bug in setup (459c9fa)

I ran into this SQL error as well
`SQLSTATE[HY000]: General error: 1364 Field 'usr_external_id' doesn't have a default value`

I think we should change this column to be nullable. 

A bigger issue is that even though the migration failed, the "Success" message is written and you cannot return to setup unless you go in and delete your config files.

glensc:
the sql error you got how? ran setup on new installation and the error happened in setup process?

the database migration is done once all config files are in place (db engine requires config files, already in previous code). so it's not a new issue.

as if i ran setup in my test it worked fine, so i wonder what's different for you?

glensc:
here's comparison of `mysqldump --extended-insert=false` of database created with previous version and this version. output is created with `diff -U 20`.

[schema.diff](https://github.com/eventum/eventum/files/977201/schema.diff.txt)

there's only difference with display width of integer columns, which i consider irrelevant, and key error of `email_account` table, which i explained earlier.

perhaps just drop the index on that table, the table is tiny, does it really need such index?

```sql
  UNIQUE KEY `ema_username` (`ema_username`,`ema_hostname`(100),`ema_folder`(100)),
```

glensc:
ok, i think i know why you got setup error and i did not. current code lets phinx itself create db connection, so the SQL mode='' is not set as in previous eventum code.

can you show what's default in your system:

```
mysql{2}> select @@sql_mode, @@version;
+------------------------+-----------------+
| @@sql_mode             | @@version       |
+------------------------+-----------------+
| NO_ENGINE_SUBSTITUTION | 5.6.32-78.1-log |
+------------------------+-----------------+
1 row in set (0.00 sec)
```

balsdorf:
```
+---------------------+----------------------------+
| @@sql_mode          | @@version                  |
+---------------------+----------------------------+
| STRICT_TRANS_TABLES | 10.0.30-MariaDB-enterprise |
+---------------------+----------------------------+
```

Reminds me I actually need to upgrade my version.

balsdorf:
For email_account table we need to keep the unique requirement, or enforce it in code. The script to download email looks at hostname, username and folder to find the script.

glensc:
ok, we need (at least for now, resolving sql mode issues is out of this PR scope) to set `SQL_MODE=''`.

so, according to [docs](http://docs.phinx.org/en/latest/commands.html#configuration-file-parameter), one can provide `"connection" => $pdo_instance`, so the whatever SQL_MODE is set by eventum code.

but this means that will (likely) lose support for fields currently in `phinx.php`, the idea of these fields was to use InnoDB or utf8mb4 in the future:

```
            'engine' => 'MyISAM',
            'charset' => 'utf8',
            'collation' => 'utf8_general_ci',
```

altho not sure how to hook SQL mode into sphinx the other way, need to investigate.

but then again, phinx is used only for database migration and we have fresh schema, maybe it's about time to fix sql mode issues at least in schema.

glensc:
- found place to set `SQL_MODE` in phinx config. 234e387d
- was able to reproduce your error

```
Database setup log:

                
Phinx by Rob Morgan - https://phinx.org. version 0.8.0

using config file ./phinx.php
using config parser php
using migration paths 
 - /home/glen/scm/eventum/eventum/db/migrations
warning no environment specified, defaulting to: production
using adapter mysql
using database e

 == 20170427170945 InitDatabase: migrating
 == 20170427170945 InitDatabase: migrated 0.3849s

 == 20170428180919 InitialData: migrating

                                                                                             
  [PDOException]                                                                             
  SQLSTATE[HY000]: General error: 1364 Field 'usr_external_id' doesn't have a default value  
                                                                                             

migrate [-c|--configuration CONFIGURATION] [-p|--parser PARSER] [-e|--environment ENVIRONMENT] [-t|--target TARGET] [-d|--date DATE]

```

glensc:
52b3241b - db setup error is now handled and you can re-submit setup form

glensc:
as for the `email_account` table. can we shorten the column names perhaps?

```
  `ema_folder` varchar(255) DEFAULT NULL,
  `ema_hostname` varchar(255) NOT NULL DEFAULT '',
  `ema_username` varchar(64) NOT NULL DEFAULT '',
```

```
mysql> select max(length(ema_username)),max(length(ema_hostname)),max(length(ema_folder)) from email_account;
+---------------------------+---------------------------+-------------------------+
| max(length(ema_username)) | max(length(ema_hostname)) | max(length(ema_folder)) |
+---------------------------+---------------------------+-------------------------+
|                        21 |                        18 |                      15 |
+---------------------------+---------------------------+-------------------------+
1 row in set (0.00 sec)
```

of course could just execute raw sql in migration file, but i don't like that:

```
$this->execute('ALTER TABLE `email_account` ADD UNIQUE KEY `ema_username` (`ema_username`,`ema_hostname`(100),`ema_folder`(100))');
```

glensc:
hmm, the max key length is limited to 1000 bytes, and table charset is utf8, i could just change the columns to be latin1:

```
            ->addColumn('ema_folder', 'string', ['null' => true, 'encoding' => 'latin1'])
            ->addColumn('ema_hostname', 'string', ['default' => '', 'encoding' => 'latin1'])
```

this way index is created

glensc:
went with latin1 columns: 882da836

glensc:
as for utf8mb4, similar key length issues appeared:

- `faq.faq_title`
- `news.nws_title`
- `user.usr_email`
- `user_alias.ual_email`

the emails change to `latin1` is sane, but not sure about the `faq` and `news` titles,
i understand that the unique keys were added to ensure no duplicate items,
i personally would drop those keys and allow duplicate titles. @balsdorf WDYT?

```diff
diff --git a/db/migrations/20170427170945_init_database.php b/db/migrations/20170427170945_init_database.php
index 3fdf311f..0106e6ba 100644
--- a/db/migrations/20170427170945_init_database.php
+++ b/db/migrations/20170427170945_init_database.php
@@ -231,7 +231,7 @@ class InitDatabase extends AbstractMigration
             ->addColumn('faq_usr_id', 'integer', ['signed' => false])
             ->addColumn('faq_created_date', 'datetime')
             ->addColumn('faq_updated_date', 'datetime', ['null' => true])
-            ->addColumn('faq_title', 'string')
+            ->addColumn('faq_title', 'string', ['encoding' => 'utf8'])
             ->addColumn('faq_message', 'text', ['length' => self::INT_REGULAR])
             ->addColumn('faq_rank', 'integer', ['length' => self::INT_TINY, 'signed' => false])
             ->addIndex(['faq_title'], ['unique' => true]);
@@ -514,7 +514,7 @@ class InitDatabase extends AbstractMigration
             ->addColumn('nws_id', 'integer', ['length' => 11, 'signed' => false])
             ->addColumn('nws_usr_id', 'integer', ['signed' => false])
             ->addColumn('nws_created_date', 'datetime')
-            ->addColumn('nws_title', 'string')
+            ->addColumn('nws_title', 'string', ['encoding' => 'utf8'])
             ->addColumn('nws_message', 'text')
             ->addColumn('nws_status', 'string', ['length' => 8, 'default' => 'active'])
             ->addIndex(['nws_title'], ['unique' => true]);
@@ -955,7 +955,7 @@ class InitDatabase extends AbstractMigration
             ->addColumn('usr_status', 'string', ['length' => 8, 'default' => 'active'])
             ->addColumn('usr_password', 'string', ['default' => ''])
             ->addColumn('usr_full_name', 'string', ['default' => ''])
-            ->addColumn('usr_email', 'string', ['default' => ''])
+            ->addColumn('usr_email', 'string', ['default' => '', 'encoding' => 'latin1'])
             ->addColumn('usr_sms_email', 'string', ['null' => true])
             ->addColumn('usr_clocked_in', 'boolean', ['default' => 0, 'null' => true])
             ->addColumn('usr_lang', 'string', ['length' => 5, 'null' => true])
@@ -970,7 +970,7 @@ class InitDatabase extends AbstractMigration
 
         $this->table('user_alias', ['id' => false])
             ->addColumn('ual_usr_id', 'integer', ['signed' => false])
-            ->addColumn('ual_email', 'string', ['null' => true])
+            ->addColumn('ual_email', 'string', ['null' => true, 'encoding' => 'latin1'])
             ->addIndex(['ual_email'], ['unique' => true])
             ->addIndex(['ual_usr_id', 'ual_email'])
         ->create();
diff --git a/phinx.php b/phinx.php
index e699d1ae..64c61421 100644
--- a/phinx.php
+++ b/phinx.php
@@ -53,12 +53,12 @@ return [
             // charset and collation must be utf8 compatible
 
             // for MySQL < 5.5.3
-            'charset' => 'utf8',
-            'collation' => 'utf8_general_ci',
+//            'charset' => 'utf8',
+//            'collation' => 'utf8_general_ci',
 
             // for MySQL >= 5.5.3
-//            'charset' => 'utf8mb4',
-//            'collation' => 'utf8mb4_unicode_ci',
+            'charset' => 'utf8mb4',
+            'collation' => 'utf8mb4_unicode_ci',
 
             // set SQL_MODE
             // http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html
```

glensc:
news and faq made utf-8 columns. ready to be merged if no more issues.

balsdorf:
Just did a fresh install and quick test. Everything looks great!

Is there any specific documentation devs need for creating new migrations or just the standard docs from phinx.org?

glensc:
i was thinking too, is there need for docs. but no, standard phinx.org docs will suffice.

however, depending whether you create tables, i suggest using our AbstractMigration, as it adds support for engine and charset that is configured in eventum project.

glensc:
you can use the `--class` or `--template` parameters to `create` command. but i just modified the created file using editor.

```
➔ phinx create --help
Usage:
  create [options] [--] <name>

Arguments:
  name                               What is the name of the migration (in CamelCase)?

Options:
  -c, --configuration=CONFIGURATION  The configuration file to load
  -p, --parser=PARSER                Parser used to read the config file. Defaults to YAML
  -t, --template=TEMPLATE            Use an alternative template
  -l, --class=CLASS                  Use a class implementing "Phinx\Migration\CreationInterface" to generate the template
...
```

glensc:
one problem has arisen, the migrations are put to root namespace:
https://github.com/robmorgan/phinx/issues/168

i propose to use `Eventum_` prefix for migration classes.

glensc:
f66f2279: the prefix is `Eventum`, it wanted CamelCase.

anyway, if you already applied previous one, you do not need to patch your `phinxlog`, as seems it compares only `version` column, the `migration_name` column does not have to match. which is probably good to know for future, and worth to know.

with both selects `phinx migrate` did not re-apply the patches, whatever the name was at the time.

```
mysql{1}> select * from phinxlog;
+----------------+----------------+---------------------+---------------------+------------+
| version        | migration_name | start_time          | end_time            | breakpoint |
+----------------+----------------+---------------------+---------------------+------------+
| 20170427170945 | InitDatabase   | 2017-05-13 01:12:43 | 2017-05-13 01:12:43 |          0 |
| 20170428180919 | InitialData    | 2017-05-13 01:12:43 | 2017-05-13 01:12:43 |          0 |
+----------------+----------------+---------------------+---------------------+------------+
2 rows in set (0.00 sec)
```

```
mysql{1}> select * from phinxlog;
+----------------+---------------------+---------------------+---------------------+------------+
| version        | migration_name      | start_time          | end_time            | breakpoint |
+----------------+---------------------+---------------------+---------------------+------------+
| 20170427170945 | EventumInitDatabase | 2017-05-13 01:14:00 | 2017-05-13 01:14:00 |          0 |
| 20170428180919 | EventumInitialData  | 2017-05-13 01:14:00 | 2017-05-13 01:14:00 |          0 |
+----------------+---------------------+---------------------+---------------------+------------+
2 rows in set (0.00 sec)
```

glensc:
it probably explains that it uses ID, considering `status` output:

```
➔ ./vendor/bin/phinx status
Phinx by Rob Morgan - https://phinx.org. version 0.8.0

using config file ./phinx.php
using config parser php
using migration paths 
 - /home/glen/scm/eventum/eventum/db/migrations
warning no environment specified, defaulting to: production
ordering by creation time

 Status  [Migration ID]  Started              Finished             Migration Name 
----------------------------------------------------------------------------------
     up  20170427170945  2017-05-13 01:14:00  2017-05-13 01:14:00  EventumInitDatabase
     up  20170428180919  2017-05-13 01:14:00  2017-05-13 01:14:00  EventumInitialData
```

glensc:
https://github.com/robmorgan/phinx/releases/tag/v0.8.0
adds new feature: "Rollbacks by start time"
https://github.com/robmorgan/phinx/pull/1045

probably should enable it, as matches our workflow. but then again it's just good to know too

Issue #234, opened 2017-01-11
PEAR mail error under PHP 7.0.8


cpinfold:
@glensc - I know you are working on removing the PEAR mail code.

In the mean time, the email portion of the system doesn't work under PHP 7.0 (under 3.11 to 3.1.9) due to a reference on:
/var/www/support/vendor/pear/mail/Mail/smtp.php on line 349

````PHP Parse error:  syntax error, unexpected 'new' (T_NEW) in /path/to/eventum/vendor/pear/mail/Mail/smtp.php on line 349````

line reads:

````$this->_smtp = &new Net_SMTP($this->host,````
should read:
````$this->_smtp = new Net_SMTP($this->host,````

I'm not sure if you would like to change the code or I can make a note in the Wiki for a PHP 7 upgrade?

glensc:
php 7 is not supported. work to replace PEAR libraries is slow work in progress.

glensc:
you may try to send the fix to pear package itself

cpinfold:
Thanks.

Someone beat me to fixing it in 2011.

Eventum's composer.json is requiring php/mail <1.30.  This issue is fixed in release 1.3.0 (the latest)

https://github.com/pear/Mail/blob/master/Mail/smtp.php#L346

1.3.0 only lists PHP 5 at the top of the file, but some work was done in Aug 2015 to make it PHP7 compatible including some test cases.

I'll report if I find other issues related to PHP7.  The limited testing I have done has worked so far.

cpinfold:
Also updated the Wiki to reflect < PHP 7.0

https://github.com/eventum/eventum/wiki/Prerequisites

glensc:
oh, i recall pear/mail lock was due some other bug. eventum abused some library internals which are no longer public in new version.

glensc:
there's another php7 incompatibility: 6ba453463

glensc:
@cpinfold could you review and perhaps test #236

glensc:
fixed with #236

laberg:
I still have this issue on only one of my servers. Even though they both have the same php version...

glensc:
this is bug report against unsupported version. also if you have a problem, you should create a new issue, not hijack random issues you think are relevant.

PR #236, opened 2017-01-12
smtp transport encapsulation

glensc:
encapsulate smtp transport use, so it could be (easily) replaced with zend mail component

fixes issues for #234
- removes internal use of PEAR Mail classes
- uses PEAR Mail 1.3.0
- removes preg_replace /e flag use 

originally i planned to replace [Mail_smtp](https://pear.php.net/package/Mail/docs/latest/Mail/Mail_smtp.html) with [Zend\Mail\Transport\Smtp](https://framework.zend.com/manual/2.4/en/modules/zend.mail.transport.html), but that can be done in a separater PR.

### DIFF
--- a/lib/eventum/class.mime_helper.php
+++ b/lib/eventum/class.mime_helper.php
@@ -177,13 +177,17 @@ class Mime_Helper
         if (self::is8bit($address)) {
             // split into name and address section
             preg_match('/(.*)<(.*)>/', $address, $matches);
+
+            $qq = preg_replace_callback(
+                '/([\x80-\xFF]|[\x21-\x2F]|[\xFC]|\[|\])/',
+                function ($m) {
+                    return "=" . strtoupper(dechex(ord(stripslashes($m[1]))));
+                },
+                $matches[1]
+            );
             $address = '=?' . APP_CHARSET . '?Q?' .
                 str_replace(' ', '_', trim(
-                    // @ because of /e:
-                    // preg_replace(): The /e modifier is deprecated, use preg_replace_callback instead
-                    // This feature was DEPRECATED in PHP 5.5.0, and REMOVED as of PHP 7.0.0.
-                    // http://www.php.net/manual/en/reference.pcre.pattern.modifiers.php
-                    @preg_replace('/([\x80-\xFF]|[\x21-\x2F]|[\xFC]|\[|\])/e', '"=" . strtoupper(dechex(ord(stripslashes("\1"))))', $matches[1])
+                    $qq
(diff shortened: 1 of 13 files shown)


lib/eventum/class.mime_helper.php
```diff
@@ -177,13 +177,17 @@ public static function encodeAddress($address)
         if (self::is8bit($address)) {
             // split into name and address section
             preg_match('/(.*)<(.*)>/', $address, $matches);
+
+            $qq = preg_replace_callback(
+                '/([\x80-\xFF]|[\x21-\x2F]|[\xFC]|\[|\])/',
+                function ($m) {
+                    return '=' . strtoupper(dechex(ord(stripslashes($m[1]))));
```
cpinfold: It took me a few minutes to figure out what this was doing.. as I don't use multi-byte characters.

@glensc Does this encoding need to be done manually?  Eg.  [mb_encode_mimeheader](http://php.net/manual/en/function.mb-encode-mimeheader.php)

Eg.
````
$str1 = '"Some Öne" <Some.One@example.org>';
$str2 = 'root@example.org';
$str3 = 'Root Mäe <root2@example.org>';

$encoded1 = mb_encode_mimeheader($str1, 'utf-8', "B");
echo "1: $encoded1\n";

$encoded2 = mb_encode_mimeheader($str2, 'utf-8');
echo "2: $encoded2\n";

$encoded3 = mb_encode_mimeheader($str3, 'utf-8');
echo "3: $encoded3\n";
````

Note that I didn't take the time to split the name and address portions which is already being done on L179. "B" Encoding is the default and could be omitted.

glensc: i didn't bother to dig into what it does. i just converted syntax. it worked so far, and idea is to drop those methods when finally able to switch to zend mail classes.

glensc: and saying "i don't use multibyte characters" is irresponsible, if anyone could send mail in they could be in any encoding; chinese spammers included :)

cpinfold: I should have said I am unfamiliar with multibyte encoding.  Spam is a good point.. fortunately haven't seen much yet.

glensc:
i'm going to merge this. if there are issues you can still report them

lib/eventum/class.mime_helper.php
glensc: also, the said regexp is not multi-byte aware. all it is capable is encode non-printable bytes. or can just say latin1 charset, which is single byte width characters.

PR #240, opened 2017-01-21
Add option to control how custom field options are ordered.

balsdorf:
Historically custom field options have been ordered alphabetically. I've now added the option to order them:
* Alphabetically / Reverse Alphabetically
* Insert / Reverse Insert
* Manually

Custom field options have also been moved to a separate page to allow for ordering and editing.

### DIFF
--- a/src/Controller/Manage/CustomFieldOptionsController.php
+++ b/src/Controller/Manage/CustomFieldOptionsController.php
@@ -1,0 +1,89 @@
+<?php
+
+/*
+ * This file is part of the Eventum (Issue Tracking System) package.
+ *
+ * @copyright (c) Eventum Team
+ * @license GNU General Public License, version 2 or later (GPL-2+)
+ *
+ * For the full copyright and license information,
+ * please see the COPYING and AUTHORS files
+ * that were distributed with this source code.
+ */
+
+namespace Eventum\Controller\Manage;
+
+use Custom_Field;
+use Eventum\Controller\Helper\MessagesHelper;
+use User;
+
+class CustomFieldOptionsController extends ManageBaseController
+{
+    /** @var string */
+    protected $tpl_name = 'manage/custom_field_options.tpl.html';
+
+    /** @var int */
+    protected $min_role = User::ROLE_ADMINISTRATOR;
+
+    /** @var string */
+    private $cat;
+
+    /** @var int */
+    private $fld_id;
+
+    /** @var array */
+    private $field_info;
+
+    /**
+     * @inheritdoc
+     */
+    protected function configure()
+    {
+        $request = $this->getRequest();
+
+        $this->fld_id = $request->request->get('fld_id') ?: $request->query->get('fld_id');
+        $this->field_info = Custom_Field::getDetails($this->fld_id);
+
+        $this->cat = $request->request->get('cat') ?: $request->query->get('cat');
+
+        if (empty($this->field_info)) {
+            $this->error(ev_gettext("Invalid custom field ID"));
+        }
+    }
+
+    /**
+     * @inheritdoc
+     */
+    protected function defaultAction()
+    {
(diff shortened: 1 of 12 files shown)


src/Controller/Manage/CustomFieldOptionsController.php
```diff
@@ -0,0 +1,89 @@
...
+    protected function configure()
+    {
+        $request = $this->getRequest();
+
+        $this->fld_id = $request->request->get('fld_id') ?: $request->query->get('fld_id');
+        $this->field_info = Custom_Field::getDetails($this->fld_id);
+
+        $this->cat = $request->request->get('cat') ?: $request->query->get('cat');
+
+        if (empty($this->field_info)) {
```
glensc: move this to "action" method

upgrade/patches/64_custom_field_option_order.sql
```diff
@@ -0,0 +1,6 @@
+/**
+ * Control the order of custom fields
+ *
+ */
+ALTER TABLE {{%custom_field_option}} ADD COLUMN cfo_rank int(10) NOT NULL DEFAULT 0;
```
glensc: avoid using WIDTH to integer columns, use just `INT`, as it's misleading, also, if you don't need negative values, use `UNSIGNED INT`. and perhaps even `UNSIGNED TINYINT` if 255 values is enough (or `SMALLINT`)

also use `AFTER` or `BEFORE` to order numeric fields first in the table.

balsdorf: Old habits die hard. Fixed.

src/Controller/Manage/CustomFieldOptionsController.php
balsdorf: I've moved this to prepareTemplate as I need it on the view page, not the update action.

upgrade/patches/64_custom_field_option_order.sql
```diff
@@ -2,5 +2,5 @@
  * Control the order of custom fields
  *
  */
-ALTER TABLE {{%custom_field_option}} ADD COLUMN cfo_rank int(10) NOT NULL DEFAULT 0;
+ALTER TABLE {{%custom_field_option}} ADD COLUMN cfo_rank INT UNSIGNED NOT NULL DEFAULT 0 AFTER cfo_fld_id;
 ALTER TABLE {{%custom_field}} ADD COLUMN fld_order_by varchar(20) NOT NULL DEFAULT 'cfo_id ASC'
```
glensc: always add some `AFTER`, `BEFORE` keyword

also, please avoid using useless commit messages like `Clean up code from code review`, such message does not explain purpose or changes of the commit. in other words it's useless later if looking at git logs what changed.

src/Controller/Manage/CustomFieldOptionsController.php
```diff
@@ -42,13 +42,8 @@ protected function configure()
         $request = $this->getRequest();
 
         $this->fld_id = $request->request->get('fld_id') ?: $request->query->get('fld_id');
-        $this->field_info = Custom_Field::getDetails($this->fld_id);
```
glensc: the class property can be removed now as well

balsdorf:
Anything else you would like seen done on this or can I merge this in?

glensc:
missing `AFTER`, `BEFORE` keywords in sql. you fixed only one query.

balsdorf:
I want that column at the end of the table so an after clause is superfluous. None of the patches after 17 use it either.

I'm fine with using BEFORE / AFTER to group ints at the start of the table or to group related fields together but in this case it is totally pointless.

glensc:
uh, this PR messed up (reset to default?) existing fields in `list.php` view?

image

balsdorf:
This PR did not touch that area. I can't reproduce the problem.

PR #247, opened 2017-03-21
Add a per project recipient flag for outgoing emails

balsdorf:


cpinfold:
Bryan,

To avoid future confusion isn't this a project _sender_ flag?  The flag is being added to the sender address, but the flag is being called a prj_recipient_type_flag in the tables.

Please correct me if I am misunderstanding this?

Craig.

- Forgive my lack of references to the code - couldn't quickly figure out how to do that.

balsdorf:
Hey Craig,

Yes that is a good point.  I based my wording on the term we already used in the global setup, but we might be better off changing it here for the future.

balsdorf:
I've just updated this PR to clarify the purpose. I've updated my new code (column and variable names) but left the old general setup values unchanged to avoid having to migrate the old values

glensc:
i don't understand the change (not enough described), but if the change is backward compatible with no changes to previous installations, i'm fine with merge.

balsdorf:
Eventum has always had the option to prepend / append a flag (such as [Eventum]) to the sender's name, e.g. "Bryan Alsdorf <issue-1234@example.com>" becomes "Bryan Alsdorf [Eventum] <issue-1234@example.com>".

Previously, this flag was set globally. This PR gives the option to override it specifically for each project.

glensc:
@balsdorf so, if no project specific value is set the global one applies? or no more global option only per project? 

i.e trying to understand what happens to existing installations after this gets merged, do they need to reconfigure something to preserve existing settings or not.

balsdorf:
If there is no project specific value the global value still applies.

Existing installations will not be impacted at all.

glensc:
upgrade/patches/65 conflict. patch from this PR should be renamed.

glensc:
"PR#247" in commit message, did not trigger update here, adding manually: de2753cff9ec95c67495e41089a01371cae9dd75

PR #248, opened 2017-04-04
replace abandoned malsup/form with jquery-form/form

glensc:
https://github.com/jquery-form/form/releases/tag/v4.0.0

> jQuery Form has a new home! The new canonical URL for the project is https://github.com/jquery-form/form.
> The major version has been incremented to 4 due to the new home of project.
> Version 4 is expected to be backward compatible with 3.51.0.

however. I can not find where do we use this library in templates.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -113,7 +113,7 @@
 		"eventum/rpc": "~4.0.0",
 		"jackmoore/autosize":"*",
 		"jasig/phpcas":"~1.3.3",
-		"malsup/form" :"*",
+		"jquery-form/form": "^4.2",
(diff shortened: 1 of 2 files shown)


balsdorf:
Dynamic Custom fields (a feature I think only I use) uses it. I'll test it.

glensc:
could you point to the code for future reference if you find it (don't forget to link to commit or tag instead of branch)

balsdorf:
https://github.com/eventum/eventum/blob/6a67e1ded7c95136c6181caf8627892b3903e6c3/htdocs/js/dynamic_custom_field.js#L149

PR #252, opened 2017-04-27
drop PEAR DB adapter support

glensc:
task from #235

however dropping table prefix and integrating phinx should be imho done in same release,
as otherwise some patches will become weird, if not even invalid.

for example running this patch [57_drop_table_prefix.php](https://github.com/eventum/eventum/blob/v3.1.10/upgrade/patches/57_drop_table_prefix.php) when there is no table prefix support in codebase, won't likely even be able to update table names and mark some patch number applied.


glensc:
@balsdorf do you see what else to do in this PR? or just let's go ahead and release 3.2.0 with new install requirements?

glensc:
that is not necessary, as if you look the diff, you see that config key to specify adapter was changed, so class default will apply.

glensc:
TODO:

- [ ] from setup: PEAR driver has 'debuginfo' property
- [ ] search sources for "mysql" and "mysqli" extensions

Issue #144, opened 2016-02-15
Bootstrap responsive template


fishmad:
Ive been customizing a bootstrap responsive version of your default html templates, wondering if you interested in me contribution it to the repository?

glensc:
yes, sure. make pull-request and then can see what to do with it (replace the default, or make multi template support)

cpinfold:
Elan - can you create a branch from master so we can all try it?

I'm about to (finally) start making some changes to remove popup's.. it may
be better to start from the bootstrap templates.

Craig.

glensc:
@cpinfold eee? i don't have any code to create branch. seems you never created any pull request? :)
github pull request is ticket asking to merge someones fork back to origin. so if there's pull request, there's always branch to clone somewhere :)

fishmad:
I went with multi template so the default remained in tack for reference.

The only core change is in **init.php** on line 60:
`$define('APP_TPL_PATH', APP_PATH . '/templates');`
change to:
`$define('APP_TPL_STYLE', 'bootstrap3'); // current options: default or bootstrap3`
`$define('APP_TPL_PATH', APP_PATH . '/templates/' . APP_TPL_STYLE);`

**and folder structure in the /templates folder was **
`templates/*allfiles.tpl.html...`
**now just looks like this:**
`templates/default/*allfiles.tpl.html...`
`templates/bootstrap3/*allfiles.tpl.html...`

any new or custom style sheets and javascripts file now exist with the root js and css folders with there own sub folders named aecatly the same as template.

**Existing way:**
- root/js
- root/css

**New way:**
- root/js/**default**
- root/js/**boostrap3**
- root/css/**default**
- root/css/**boostrap3**

**Sample work in progress, you have lots of files I need to change.**
![image](https://cloud.githubusercontent.com/assets/12012101/13101328/b57d853c-d599-11e5-91cb-7b60792bf7a9.png)
![image](https://cloud.githubusercontent.com/assets/12012101/13101320/99d9f36a-d599-11e5-91b9-c0688c767e92.png)
![image](https://cloud.githubusercontent.com/assets/12012101/13101352/eef28510-d599-11e5-825a-c61386bbf04a.png)
![image](https://cloud.githubusercontent.com/assets/12012101/13101356/f62ca5d6-d599-11e5-93fe-30788c9b0def.png)
![image](https://cloud.githubusercontent.com/assets/12012101/13101366/11c1ae90-d59a-11e5-9157-01eb145febcf.png)
[...]

PR #224, opened 2016-12-01
ADD icons for priority

phavel:
Proposal icon for priority.

Icons: Farm-fresh http://www.fatcow.com/free-icons
DB: add "pri_icon" column in priority table

list
view
manage-1
manage-2

PR #253, opened 2017-05-03
use font awesome

glensc:
The iconic font and CSS framework
Font Awesome is a full suite of 675 pictographic icons for easy scalable vector graphics on websites.

benefits:
- modern icons
- hdri/retina/etc support
image
image

- ***fill me***
- #144 seems to be dead, so at least this change should be simple

seems easy to use: http://fontawesome.io/examples/

should start mapping ui where to make replacements:
- [x] rank order arrows: [arrow-up](http://fontawesome.io/icon/arrow-up/), [arrow-down](http://fontawesome.io/icon/arrow-down/) or any other [#directional icon](http://fontawesome.io/icons/#directional)
image
- [x] help icon:
image
- [x] upload spinner 
image
- [x] attachment icon
image
- [x] reply icons
image
- [ ] ....

this can be image names, css class names, or dom div names, or whatever identifies what needs to be replaced


glensc:
only these images left:
- htdocs/images/eventum.gif - eventum logo. not used, but could be used in the future 
- htdocs/images/no_data.gif - used by stats.php
- htdocs/images/priority/*.png - icons from #224, could be replaced with FA versions, if so, separate PR after this one.

Issue #251, opened 2017-04-19
Move attachments out of the database


balsdorf:
For large installations, keeping a large number of attachments in the database can hinder database performance and create logistical problems. I plan on changing Eventum to (optionally?) store files on the filesystem. This creates alternative logistical challenges of course but I believe it is worth it.

Open questions:
* Should Eventum have the option to keep attachments in the database?
* What directory should attachments be stored in?

Thoughts?

glensc:
i would leave it optional, because i can do the migration later, not block update of eventum because attachment migration not completed.

about directory, new setup.php option (not the php constant), that defaults to `var/attachments`?
http://symfony.com/doc/current/quick_tour/the_architecture.html

i.e the directory should definately not be public, as need to acl access to attachments.

and as for implementation, perhaps implement storage driver subsystem, then could store attachments remotely, like aws s3, or [minio](https://www.minio.io/)

glensc:
about storage drivers, quick find lead me to [league/flysystem](http://flysystem.thephpleague.com/), but definately there are others.

we just need to implement compatible Provider for our current Database mode.

glensc:
created branch: https://github.com/eventum/eventum/pull/254

new dependencies: none, but suggests other packages:
```
Package operations: 1 install, 0 updates, 0 removals
  - Installing league/flysystem (1.0.28): Loading from cache
league/flysystem suggests installing league/flysystem-aws-s3-v2 (Allows you to use S3 storage with AWS SDK v2)
league/flysystem suggests installing league/flysystem-aws-s3-v3 (Allows you to use S3 storage with AWS SDK v3)
league/flysystem suggests installing league/flysystem-azure (Allows you to use Windows Azure Blob storage)
league/flysystem suggests installing league/flysystem-cached-adapter (Flysystem adapter decorator for metadata caching)
league/flysystem suggests installing league/flysystem-copy (Allows you to use Copy.com storage)
league/flysystem suggests installing league/flysystem-dropbox (Allows you to use Dropbox storage)
league/flysystem suggests installing league/flysystem-eventable-filesystem (Allows you to use EventableFilesystem)
league/flysystem suggests installing league/flysystem-rackspace (Allows you to use Rackspace Cloud Files)
league/flysystem suggests installing league/flysystem-sftp (Allows you to use SFTP server storage via phpseclib)
league/flysystem suggests installing league/flysystem-webdav (Allows you to use WebDAV storage)
[...]

PR #254, opened 2017-05-04
Use league/flysystem to abstract attachments

glensc:
Move attachments out of the database #251

uses [league/flysystem](http://flysystem.thephpleague.com/), [requires php 5.5.9+](https://github.com/thephpleague/flysystem/blob/1.0.40/composer.json#L17)


glensc:
mount manager: http://flysystem.thephpleague.com/mount-manager/ could came useful

glensc:
may even decide to use caching: http://flysystem.thephpleague.com/caching/

glensc:
so the idea is to abstract `issue_attachment_file` table access, `issue_attachment` can stay in database.

- [ ] create `src/Flysystem/Adapter/AttachmentStore.php` using current database (better name suggestion welcome)
- [ ] use [MountManager](http://flysystem.thephpleague.com/mount-manager/)  and map `attachment://` to configured adapter
- [ ] use calls like `$filesystem->has("attachment://$iaf_id")`, `$filesystem->put($path, $contents)`, `$filesystem->read($path)`

@balsdorf WDYT?

glensc:
probably should just store attachments as `$iaf_iat_id-$iaf_id-$iaf_filename` because there isn't enough context information for anything else. altho could use subdirs. or leave the decision to user, i.e have it configurable?

```
mysql> desc issue_attachment_file;
+------------------+------------------+------+-----+---------------------+----------------+
| Field            | Type             | Null | Key | Default             | Extra          |
+------------------+------------------+------+-----+---------------------+----------------+
| iaf_id           | int(10) unsigned | NO   | PRI | NULL                | auto_increment |
| iaf_iat_id       | int(10) unsigned | NO   | MUL | 0                   |                |
| iaf_file         | longblob         | YES  |     | NULL                |                |
| iaf_filename     | varchar(255)     | NO   |     |                     |                |
| iaf_filetype     | varchar(255)     | YES  |     | NULL                |                |
| iaf_filesize     | varchar(32)      | NO   |     |                     |                |
| iaf_created_date | datetime         | NO   |     | 0000-00-00 00:00:00 |                |
+------------------+------------------+------+-----+---------------------+----------------+
7 rows in set (0.01 sec)
```

glensc:
searched github for some existing adapters, mysql, pdo flysystem

- https://github.com/CrazyDevil/flysystem-mysql - GPL, 6 releases, 12 commits, looks incomplete what hardcoded table structure
- https://github.com/IntegralSoftware/flysystem-pdo-adapter - MIT, 2 releases, 5 commits, generic pdo adapter, looks usable, but also fixed db structure
- https://github.com/phlib/flysystem-pdo - LGPL, 3 releases, 43 commits, uses chunks, compression, table_prefix configurable, memory usage optimizations

looks to me phlib/flysystem-pdo is a win

glensc:
phlib/flysystem-pdo changes past last release include [drop php 5.5 support](https://github.com/phlib/flysystem-pdo/commit/8e6893dac34d8a2787563766144ad2d92407f120)

we can perhaps also up min requirement to php 5.6, which leaves only 5.6, as 7.0 is not supported for eventum :)

it also includes [change license to LGPL-3](https://github.com/phlib/flysystem-pdo/commit/c92a220528e78ead4689dc4476b5985a69c25b09), that's still ok with eventum right?

glensc:
[phlib/flysystem-pdo](https://github.com/phlib/flysystem-pdo) seems usable. to simplify integration, should use it's [provided table structure](https://github.com/phlib/flysystem-pdo/blob/0.0.3/schema/mysql.sql). to use for multiple purposes (besides attachments), we can play with "table_prefix". in this example i used default "flysystem", so that could be "attachment" for this use case.

```
mysql> select * from flysystem_path;
+---------+------+---------------------------+------------+------------+------+---------------+---------------------+
| path_id | type | path                      | mimetype   | visibility | size | is_compressed | update_ts           |
+---------+------+---------------------------+------------+------------+------+---------------+---------------------+
|       1 | file | testPhlibFlysystemPdo.txt | text/plain | public     |   23 |             0 | 2017-05-12 23:11:47 |
+---------+------+---------------------------+------------+------------+------+---------------+---------------------+
1 row in set (0.00 sec)

mysql> select * from flysystem_chunk;
+---------+----------+-------------------------+
| path_id | chunk_no | content                 |
+---------+----------+-------------------------+
|       1 |        0 | bite my shiny metal ass |
+---------+----------+-------------------------+
1 row in set (0.00 sec)
```

balsdorf:
I'm good with upping requirement to php 5.6.

License compatibility is a good question. Eventum is licenses as "GNU General Public License, version 2 or later (GPL-2+)". GPL2 is incompatible with LGPL3, but since it says "or later" you can deem it GPL3 which then means it is compatible with LGPL3. https://www.gnu.org/licenses/gpl-faq.html#AllCompatibility

balsdorf:
The issue with switching to using this adapter (which looks nice) is that all our current attachments in the DB will have to be migrated to the new table, which could be a huge job.

balsdorf:
For the filesystem, I like subdirectories and I'd use one of these:

* $prj_id/$iss_id/$iat_id-$iaf_id.$extension
* $prj_id/$iss_id/$iat_id-$iaf_id-$filename.$extension

balsdorf:
So I created a "LegacyAdapter" to read from our current, attachment stored files. From messing around with this, I think we should use MountManager to support multiple adapters at the same time. For the most part, I think we would just use `legacy://' and either 'pdo://' or 'local://' but the setup file could be used to be specify additional adapters. 

We would still use issue_attachment and issue_attachment_file with the addition of a iaf_flysystem_path field. We would continue to store meta data ourselves since we need to easily display it, but reading / writing the actual contents of the files would all be handled by flysystem. 

Old files would be read from `legacy://` but all new files would be written to the default adapter.

We could also provide a migrate command to move files from legacy to the default adapter, but for users with large databases this could be quite painful.

Thoughts?

glensc:
@balsdorf rebased against master to resolve conflicts. go ahead and push your changes out. your plan is probably fine.

balsdorf:
So I've just pushed my work on this. At first I tried to make minimal changes but wasn't happy with how messy it was so I pretty much redid all of the attachment system. Here are the key points.

New class `Eventum\Attachment\StorageManager`:
This class abstracts the details of Flysystem away and is responsible for loading the adapters from the config file. This can also be used if we want to store non-attachment files (not sure what yet, but maybe a logo for customization css?). I've thought that this maybe should be in `Eventum\Storage\`

New class `Eventum\Attachment\AttachmentManager`:
Handles most of the functionality in the old attachment class. Handles uploads, associates them with issues, lists attachments, etc.

New class `Eventum\Attachment\Attachment`:
Represents an individual attachment in the database. Saves / Loads meta data from `issue_attachment_file`.

New class `Eventum\Attachment\AttachmentGroup`:
Represents a group of attachments related to an issues. Maps to the `issue_attachment` table.

Adapter for old files in `Eventum\Attachment\EventumLegacyAdapter`:
Loads data from issue_attachment_file. Not capable of writing or updating. Possibly should be moved to `Eventum\Storage\`

Since I was re-writing this I also added the ability to specify a minimum role for `AttachmentGroups` instead of just public/private.

Currently, by default attachments are still stored in the database, just with the new PDO Adapter. I believe we should make this configurable in the admin and during installation. For the local adapter I used the subpath "storage/" but I'm open to better suggestions. 

As far as paths go, when files are first uploaded (before they are associated with an issue) the path is "unassociated/$iaf_id-$filename". When they are associated they are moved to "$issue_id/$iaf_id-$filename".


TODO:
- [ ] Management page for configuring Storage
- [ ] Choose between Filesystem and DB storage on install
- [ ] Update workflow system for new Attachment objects
- [ ] Script to move from one backend to another
- [ ] Display nice error if attachment cannot be found in any matching backends.

balsdorf:
@glensc Do you think we should default to storing attachments in the database or the filesystem for new installs?

glensc:
i would leave by default in database, easier to setup. and admin interface thus not needed. could stay as advanced setup.

balsdorf:
Any more features or fixes you would like on this? I'm running through some more testing but I'm happy with how things look.

glensc:
i haven't looked at this more than a code. should at least try to test update on dev machine.

ps: remove WIP (work in progress) from PR title when it's no longer WIP :)

glensc:
Argument packing is available since PHP 5.6, see `StorageManager`. our current min requirement is 5.5

balsdorf:
Earlier in this thread we discussed that we would have to bump the min level to 5.6:
https://github.com/eventum/eventum/pull/254#issuecomment-301165654

glensc:
legacyadapter seems broken: getMimetype, getTimestamp wrap getMetadata, but should return some property not whole array?

glensc:
migrate script: does it actually move, or just copies from one to other storage? can the process be interrupted and resumed?

perhaps make the script to do copy and second operation (flag? second script?) to do delete if already migrated?

this will use more resources (disk, time), but is more safer to execute. would be sad if disk got full and some attachment is lingering between two worlds, or something similar. besides mysql space is not reclaimed until one executes `OPTIMIZE TABLE`, at least for MyISAM, perhaps print such message at the end of the script.

glensc:
ok, update config.platform to 5.6 in composer.json, maybe some other places. we should target this to 3.3.0 release then. (first release whatever in master as 3.2.2) and wiki docs too.

balsdorf:
I know it doesn't make sense, but getMimeType() and getTimeStamp() are supposed to return the whole array of meta data. At last that is what the [LocalAdapter](https://github.com/thephpleague/flysystem/blob/0b657b1580dbb063c1fe979e693dd8e7a1f6883e/src/Adapter/Local.php#L308-L327) does.

The migrate script uses the [mount manager ](https://github.com/thephpleague/flysystem/blob/0b657b1580dbb063c1fe979e693dd8e7a1f6883e/src/MountManager.php#L254)to move the file. Internally, that code copies the file and once the copy is confirmed deletes the old file. The only risk would be if for some reason the path in eventum cannot be updated. I do believe this is a small risk though and would only impact one file before erroring out. Considering this is an advanced script I think that is an acceptable risk. However, I did find one small bug I will fix and will also add some warnings to it and a note about OPTIMIZE TABLE.

balsdorf:
Thanks for those fixes BTW. I'll update the references to 5.6 in the code and update the wiki

glensc:
some nitpicking i considered:

perhaps use [touch](http://php.net/touch) to set filesystem file dates from database. could be useful for some cleanups or analyze tools that use file timestamps.

as i understood flysystem stores such meta in it's database, but be useful to duplicate this on actual files as well. applies to migrate script mostly, but may have effect when move is used elsewhere.

glensc:
note to self:
this PR removes `iat_status` column (`20170630200513_eventum_attachments_drop_status.php`), so going back from this branch (say master), makes whole eventum spew errors. 

```mysql
alter table issue_attachment add
  `iat_status` enum('internal','public') NOT NULL DEFAULT 'public';
```

glensc:
@balsdorf i also wish you do the last  3.2 release, to see the validity of docs and maybe some ideas/issues you encounter! :)

balsdorf:
Good idea on touching the file, I've incorporated that into my changes to migration script.

glensc:
i probably want to relocate `APP_PATH . '/var/storage/'` -> `/var/lib/eventum/storage` when i deploy this using rpm package. should we setup dynamic value for this using Setup or I should just patch every match for direct use in codebase?

balsdorf:
I was thinking about using a dynamic value from setup, but I didn't want
someone to change it  and think all the files would be moved automatically.
However, I'll add that  as it is better then you having to patch a bunch of
files.

balsdorf:
When going to implement a path in setup for you I realized that no changes were needed, you just need to specify your own adapter:
```
  'attachments' =>
  array (
    'default_adapter' => 'local',
    'adapters' => array(
        'local' =>  array(
            'class' =>  '\\League\\Flysystem\\Adapter\\Local',
            'options'   =>  array('/path/to/my/system/')
        ),
    ),
  ),
```

This will override the "default" local file location. The only item you would have to patch is in the [migration script](https://github.com/eventum/eventum/blob/issue-251-attachmentstore/contrib/migrate_storage_adapter.php#L100), and only if you want to "touch" the files

glensc:
yay for merge! don't forget to update docs for 5.6 min requirement. thanks!

wish me good luck rebasing #263 (it contained also changes (cleanups)) to attachment area :(

glensc:
@balsdorf found some error, the `$mimetype` is always uninitialized here:
[lib/eventum/class.misc.php:740](https://github.com/eventum/eventum/blob/b148c51b7cc76b7289302cab57bfe5a0896b8f01/lib/eventum/class.misc.php#L740)

PR #257, opened 2017-05-07
add ExtensionLoader class

glensc:
ExtensionLoader encapsulates repeated code for loading "backends".

changed also how classes are listed, abstract classes are automatically skipped.

also crm had different naming convention, perhaps allow such breakage and unify it's filename to classname mapping with other backends?

i have no examples how the crm files should be named (documentation did not explain it either).

also crm has such hack, perhaps it could be dropped by defining new standard and naming convention:

```php
            $list['class.' . $file . '.php'] = $classname;
```

maybe also move them to namespace `Eventum\CRM` under psr-1:
- lib/eventum/crm/* -> src/CRM/

if so, i'd do that in separate PR

### DIFF
--- a/lib/eventum/class.crm.php
+++ b/lib/eventum/class.crm.php
@@ -304,8 +304,8 @@ abstract class CRM
             APP_LOCAL_PATH . '/crm',
         ];
 
-        $extensionLoader = new ExtensionLoader();
-        $files = $extensionLoader->getFileList($dirs);
+        $extensionLoader = new ExtensionLoader($dirs);
+        $files = $extensionLoader->getFileList();
 
         $list = [];
         foreach ($files as $file => $classname) {
--- a/lib/eventum/class.crm.php
+++ b/lib/eventum/class.crm.php
@@ -299,14 +299,7 @@ abstract class CRM
      */
     public static function getBackendList()
     {
-        $files = static::getExtensionLoader()->getFileList();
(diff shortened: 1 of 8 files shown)


balsdorf:
Nice, I'll check it out and look at how best to handle CRM.  Either namespaces or extending ExtensionLoader to be able to check if a class is a subclass of CRM

glensc:
as for future plans, i have such idea: #259

that would make placement of class file locations rather irrelevant.

balsdorf:
The old directory / naming convention for CRM is:
```
crm/
      blah/
             class.blah.php
             related files & subdirectories for blah CRM
```

The actual class name would be "Blah" and it has to inherit from CRM.

I've changed it so "class.blah.php" now has to be directly in lib/crm to greatly simplify the code instead of recursively searching subdirectories. Considering I'm probably the only person using CRM I'm okay with it.

glensc:
1f25ebd: Due to [PHP Bug #53727](https://bugs.php.net/bug.php?id=53727), `is_subclass_of` might return inconsistent results on some PHP versions if `$this->parent_class` can be an interface. If so, you could instead use `ReflectionClass::implementsInterface`.

glensc:
the partner backend loading seems to be broken, found it out when tried to update in #259. will fix it hopefully in #259.

PR #259, opened 2017-05-08
add extension support for workflows/customfields/partners/crm

glensc:
The idea is that in setup.php you define path to your extension project.

- Eventum Core will walk over the list
- create class instances
- call registerAutoloader($autoloader) on those classes

in admin views, or places asking for list of workflow classes
- Eventum Core calls `getAvailableWorkflows` and merges the list

when specific workflow is saved in admin, the implementing class would be stored in database (opposed to currently class filename).

so the change is that Workflow/Custom_Field/Partner/CRM can be implemented any class that can be autoloaded. the classes can be available by composer.json or via the `registerAutoload` method.

----
Add new extension to eventum core, add to `config/setup.php` `extensions` entry and implement `ExtensionInterface` methods. `AbstractExtension` is provided which should be extended.

```php
return [
    'extensions' => [
        'Eventum\Extension\ExtensionExample' => '/path/to/ExtensionExample.php',
    ],
```

Example extension is provided:
- [docs/examples/extension/ExampleExtension.php](https://github.com/eventum/eventum/blob/f1d0a5424f1bb9eb953ce77fe65ace1ea7accc4e/docs/examples/extension/ExampleExtension.php)

The structure how to lay out classes is irrelevant as long as the classes can be autoloaded. autoloader can be configured by `registerAutoload` method.

NOTE: the existing layout and loading by filename is still preserved.

### DIFF
--- a/db/migrations/20170512210935_eventum_extension_class.php
+++ b/db/migrations/20170512210935_eventum_extension_class.php
@@ -26,6 +26,8 @@ class EventumExtensionClass extends AbstractMigration
         $this->modifyColumn('custom_field', 'fld_backend', ['null' => true]);
         $this->modifyColumn('project', 'prj_customer_backend', ['null' => true]);
         $this->modifyColumn('project', 'prj_workflow_backend', ['null' => true]);
+        $this->modifyColumn('user', 'usr_par_code', ['null' => true]);
+        $this->modifyColumn('issue_partner', 'ipa_par_code');
     }
 
     private function modifyColumn($table, $column, $options = [])
--- a/db/migrations/20170512210935_eventum_extension_class.php
+++ b/db/migrations/20170512210935_eventum_extension_class.php
@@ -11,27 +11,27 @@
  * that were distributed with this source code.
  */
 
-use Phinx\Migration\AbstractMigration;
+use Eventum\Db\AbstractMigration;
 
 class EventumExtensionClass extends AbstractMigration
 {
     /**
-     * change() doesn't support MODIFY statement
-     * there's no down() because we just increase width here
+     * changeColumn() doesn't support down(),
+     * and we don't implement it ourselves,
+     * because it's just width increase change.
      */
     public function up()
     {
-        $notNullable = 'VARCHAR(255) CHARACTER SET ASCII NOT NULL';
-        $nullable = 'VARCHAR(255) CHARACTER SET ASCII DEFAULT NULL';
-
-        $this->modifyColumn('partner_project', 'pap_par_code', $notNullable);
-        $this->modifyColumn('custom_field', 'fld_backend', $nullable);
-        $this->modifyColumn('project', 'prj_customer_backend', $nullable);
-        $this->modifyColumn('project', 'prj_workflow_backend', $nullable);
+        $this->modifyColumn('partner_project', 'pap_par_code');
+        $this->modifyColumn('custom_field', 'fld_backend', ['null' => true]);
+        $this->modifyColumn('project', 'prj_customer_backend', ['null' => true]);
+        $this->modifyColumn('project', 'prj_workflow_backend', ['null' => true]);
     }
 
-    private function modifyColumn($table, $column, $definition)
+    private function modifyColumn($table, $column, $options = [])
     {
-        return $this->execute("ALTER TABLE `{$table}` MODIFY `{$column}` {$definition}");
+        $options += ['limit' => self::TEXT_TINY, 'encoding' => 'ascii'];
+        $this->table($table)
+            ->changeColumn($column, 'string', $options);
     }
 }
--- a/db/migrations/20170512210935_eventum_extension_class.php
+++ b/db/migrations/20170512210935_eventum_extension_class.php
@@ -1,0 +1,37 @@
+<?php
+
+/*
+ * This file is part of the Eventum (Issue Tracking System) package.
+ *
+ * @copyright (c) Eventum Team
+ * @license GNU General Public License, version 2 or later (GPL-2+)
+ *
+ * For the full copyright and license information,
+ * please see the COPYING and AUTHORS files
+ * that were distributed with this source code.
+ */
+
+use Phinx\Migration\AbstractMigration;
+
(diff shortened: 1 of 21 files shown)


glensc:
later on the `Extension` class could provide more methods, like:

```
    public function getSubscribedEvents()
    {
        return [
            WorkflowEvents::EMAIL_BLOCKED => 'handleBlockedEmail',
        ];
    }
```

glensc:
the config should likely be Classname => Filename, as just Filename gives the old question back, what is the class name inside the file...

```php

return [
    'extensions' => [
        'Eventum\Extension\ExtensionExample' => '/home/glen/scm/eventum/eventum/src/Extension/ExtensionExample.php',
    ],
```

glensc:
not handled here: templatedir. can be made as separate PR

glensc:
@balsdorf i'm a bit stuck with Partners part of the change. can you give example of partner class names?

i.e what is stored as  `$par_code` in database?

balsdorf:
@glensc 

`class Acme_Partner_Backend extends Abstract_Partner_Backend` would be in class.acme.php and $par_code would be "acme".

glensc:
so while other backends store in database filename, i.e `class.acme.php`, partners store just `$par_code`?

also, other backends do not show the database field anywhere in ui (excluding admin here), but partners do? do we need to add `getCode()` method to `Abstract_Partner` if wanting to store class name in database (that is `Acme_Partner_Backend` in this example)?

balsdorf:
Correct.

I realize it isn't consistent, but if we are going to make it consistent I
think changing the other backends to store a code instead a filename is
cleaner.

Looking around it seems that par_code is not displayed anywhere, we call
$backend->getName() anytime the partner needs to be displayed.
Abstract_Partner already has getCode() implemented but it looks like it is
never called.

glensc:
it allows you just use single file configuration, everything else is just class names, thus you are free to use your own namespace, naming convention, psr-4.

however for other backends, my intention was to store classname in database. currently it's able to resolve class.foo.php to class name and the instance (via ExtensionLoader) and just full classname using the new way.

glensc:
@balsdorf all four backend types should work at least the new way. partner/crm old way may be broken, i tested them as good as i could.

altho i just pushed changes, so there may be new bugs :)

glensc:
as about "overly verbose statement", i'm not sure too if Interface is necessary, currently looks like copy paste of Abstract class, and Abstract class is what is actually used when creating external Extension.

perhaps it's just useful for defining internal clean api :)

balsdorf:
I've found a few bugs relating to partners. One I fixed in 20b4c99 (just a display issue). The second one is pap_par_code is a varchar(30), which is too short if you are using namespaced code. How long do you think we should allow? 100? 200? 255?

glensc:
I'll go with `varchar(255) latin1`, the db part is not handled anyway yet. as ideally should convert values to classname instead of current code/filename. but could do the field in later migration file? or you prefer to do it in same migration?

balsdorf:
I don't have an opinion on if it should be one migration or two but varchar(255) latin1 sounds good.

glensc:
ok, added db data migration as well. did only testing on partner_project, the others assumed will migrate the same way.

so, it's ready for merge from my side.

glensc:
or given, that the db is migrated, there's no point having support for previously loading by filename? uh. no. probably should remove the db migration and postpone it to 3.3.0 as existing installations will break as no class path is setup.

glensc:
so, either the db should be migrated, or admin be added support to show (convert?) also currently saved db value, otherwise just editing project or custom_field resets the existing value.

glensc:
removed the db update migration from this branch, moved to new branch:
https://github.com/eventum/eventum/commits/extension-framework-db-migrate

glensc:
third option, provide builtin extension that is able to load classes from legacy paths.

glensc:
@balsdorf found some dead code i think:

[lib/eventum/class.custom_field.php:1051-1065](https://github.com/eventum/eventum/blob/v3.1.10/lib/eventum/class.custom_field.php#L1051-L1065)
`$row['field_options']` is always overwritten on line `1064`,
so basically [lines 1053-1060](https://github.com/eventum/eventum/blob/v3.1.10/lib/eventum/class.custom_field.php#L1053-L1060) can just be dropped.

can there be side effects, like calling those fills something?

glensc:
@balsdorf leaving this PR alone for a while. seems complete to me.

glensc:
@balsdorf what is `user.usr_par_code`, it needs replaced too?

balsdorf:
Yeah I think that code was safe to remove.

user.usr_par_code is used to set a partner for a user. This user will be restricted to access issues assigned to that partner. Similar situation for issue_partner.ipa_par_code, it maps issues to partners. So yes they will need to be expanded.

Besides that, my tests all looks good.

glensc:
added the two fields to conversion: 3a075a866

```diff
@@ -790,7 +790,7 @@
 /*!40101 SET character_set_client = utf8 */;
 CREATE TABLE `issue_partner` (
   `ipa_iss_id` int(11) unsigned NOT NULL,
-  `ipa_par_code` varchar(30) NOT NULL,
+  `ipa_par_code` varchar(255) CHARACTER SET ascii NOT NULL,
   `ipa_created_date` datetime NOT NULL,
   PRIMARY KEY (`ipa_iss_id`,`ipa_par_code`)
 ) ENGINE=MyISAM DEFAULT CHARSET=utf8;
@@ -1847,7 +1847,7 @@
   `usr_last_login` datetime DEFAULT NULL,
   `usr_last_failed_login` datetime DEFAULT NULL,
   `usr_failed_logins` int(11) unsigned NOT NULL DEFAULT '0',
-  `usr_par_code` varchar(30) DEFAULT NULL,
+  `usr_par_code` varchar(255) CHARACTER SET ascii DEFAULT NULL,
   PRIMARY KEY (`usr_id`),
   UNIQUE KEY `usr_email` (`usr_email`)
 ) ENGINE=MyISAM AUTO_INCREMENT=851 DEFAULT CHARSET=utf8;
```

PR #256, opened 2017-05-05
drop Mail_rfc822 PEAR Mail requirement

glensc:
the changes are done, but they need good testing, as i had to assume what some methods return. i.e RAW or QP-encoded. and the results are different, but still valid. however if eventum combines result by it's own then the statement is no longer true.

PR #263, opened 2017-05-11
drop pear/mime_decode, pear/mail_mime

glensc:
to address issues from #256, it's about time to drop pear/mime_decode, pear/mail_mime
most of the needed the functionality is already present in `MailMessage` class that uses Zend/Mail.

need to rip out uses of `Mime_Helper::decode` which uses the offending module. surprisingly the list is not that huge.

- [x] Mail_Helper::getMessageID
- [x] Mail_Helper::saveOutgoingEmailCopy
- [x] Mail_Queue::_getHeaders
- [x] Mime_Helper::getAttachment
- [x] Mime_Helper::getAttachmentCIDs
- [x] Mime_Helper::getAttachments
- [x] Mime_Helper::getCharacterSet
- [x] Mime_Helper::hasAttachments
- [x] Note::convertNote
- [x] Routing::route
- [x] Routing::route_drafts
- [x] Routing::route_emails
- [x] Routing::route_notes
- [x] Support::associate
- [x] Support::extractAttachments
- [x] Support::getEmailInfo :smiling_imp: 
- [x] Support::moveEmail
- [x] Support::sendEmail
- [x] Mail_Helper::rewriteThreadingHeaders

also:
- [x] drop use of pear/mail_mime
- [x] drop use of pear/mail_mimedecode

workflow changes:
```diff
-    public function handleNewEmail($prj_id, $issue_id, $message, $row = null, $closing = false)
+    public function handleNewEmail($prj_id, $issue_id, MailMessage $mail, $row, $closing = false)

-    public function preEmailDownload($prj_id, $info, $mbox, $num, &$message, $email, $structure = null)
+    public function preEmailDownload($prj_id, ImapMessage $mail)
```

NOTE: when testing out this branch, be sure to run `composer update`.


glensc:
haha, @balsdorf. found some bug from 4ba84c91 decade ago

it adds sender email as message-id header in [Issue::close()](https://github.com/eventum/eventum/blob/4ba84c91f1ee542fc9d3c90e9db1c0c0954fb633/eventum/include/class.issue.php#L1254):
```php
   $message_id = User::getFromHeader($usr_id);
```

and indeed, my db contains quite some of such entries :)

```
select count(*) from support_email where sup_message_id !='' and sup_message_id not like '<%';
+----------+
| count(*) |
+----------+
|    19437 |
+----------+
1 row in set (0.19 sec)
```

i'm going to omit the message-id creation, as current `MailMessage` generates missing message-id itself.

glensc:
found some rather weird method. and of course the phpdoc useless

- [Support::getRecipientsCC($cc)](https://github.com/eventum/eventum/blob/v3.2.1/lib/eventum/class.support.php#L1935-L1951)

it seems to convert commas to semicolons and then split. i'm not sure what is the input, but comma is valid value (if quoted) in email "name" part.

balsdorf:
Yeah it isn't clearly named. Seems to be used to split / sanitize any input in the cc field when sending an email. And like a lot of the old mail handling functionality it is semi broken.

glensc:
dropped some methods, cooked (probably) compatible api for workflow methods to still use mime_decode-like object. however also dropped unused `Notification::isBounceMessage`, but my workflow method used it.

i don't really like that workflow bridge, @balsdorf what you think?

balsdorf:
I think we might be better just making a backwards incompatible change and passing the $mail object around.  Those workflow methods were pretty ugly, passing $structure and $row, or everything individually.

If we just pass one object in and out it will be easier to work with and easier to maintain in the future.

glensc:
the new implementation will fail to process broken mails at all. i don't know what the previous pear library did. it's because raw input is loaded to Zend\Mail structures and if that fails, exception is thrown.

examples of broken emails:
- empty content-type for attachments
- content-type has name="foo" attribute where "foo" is non-ascii (not encoded)
- truncated email, i.e lack of closing multipart separator

altho my examples are rather old (from 2005 so far).

not sure what to think of this. probably good if broken emails don't reach system as they are likely spammers or broken mail clients?

balsdorf:
The Pear library probably tried to keep going and then failed unpredictably.

I think having broken emails fail is good, as long as it doesn't mistakenly fail for valid emails.

glensc:
https://github.com/eventum/eventum/pull/263#issuecomment-309513488

so, it's fine to you to drop all code that i encounter that uses semicolons as recipients separator?

well. only place where it could be allowed is `POST` data handling in `Controller` classes, but never inside `src/*` or `lib/*` classes. i.e to allow user to enter recipients separated by semicolon. but even that should be handled then by UI in javascript. like we have user picker in assignment or notification list

balsdorf:
Yes. Lets get rid of the old ugly code. Since you are doing this we should do it right.

glensc:
found bug that some workflow backend methods were declared as static, likely copy paste from workflow class. likely nobody implements these in their backends (even you @balsdorf ?) that it hasn't been noticed so far.

```diff
@@ -692,7 +689,7 @@ class Abstract_Workflow_Backend
      * @param $new_prj_id integer
      * @since 3.1.7
      */
-    public static function handleIssueMovedFromProject($prj_id, $issue_id, $new_prj_id)
+    public function handleIssueMovedFromProject($prj_id, $issue_id, $new_prj_id)
     {
     }

@@ -704,7 +701,7 @@ class Abstract_Workflow_Backend
      * @param $old_prj_id integer
      * @since 3.1.7
      */
-    public static function handleIssueMovedToProject($prj_id, $issue_id, $old_prj_id)
+    public function handleIssueMovedToProject($prj_id, $issue_id, $old_prj_id)
     {
     }

@@ -718,7 +715,7 @@ class Abstract_Workflow_Backend
      * @return array A key/value array with the keys being field names in the issue table
      * @since 3.1.7
      */
-    public static function getMovedIssueMapping($prj_id, $issue_id, $mapping, $old_prj_id)
+    public function getMovedIssueMapping($prj_id, $issue_id, $mapping, $old_prj_id)
     {
         return $mapping;
     }
```

glensc:
`pear/mail_mime` and `pear/mail_mimedecode` are both dropped, workflow methods also updated.

now it's needed **heavy** testing. i've tested so far only ui part, haven't ran crons yet on the new code.

balsdorf:
Yeah copy and paste error. I actually do use those methods and they are implemented as static. However I don't think PHP complains if you call a static method non statically so I never noticed. Thanks for catching.

glensc:
My PHP did, when i had implemented one of the methods in my workflow. that is how i noticed. probably  you have your php error reporting turned down that you do not notice.

glensc:
https://github.com/eventum/eventum/pull/263#issuecomment-317176271
cherry-picked to master: 72cb6602d

glensc:
merging. lets get this out moving

Issue #192, opened 2016-09-09
Datatables error on Manage Users page


slay123:
Open Manage Users page (https://eventum.example.org/manage/users.php) and enjoy the error:
![screenshot 2016-09-09 18 40 10](https://cloud.githubusercontent.com/assets/12842560/18412400/304f161a-7796-11e6-9df0-f6d6c0456ab0.png)

At least, when clicking okay, list will be loaded etc. Probably DT is loaded twice.

glensc:
Unable to reproduce in clean dev env. @balsdorf what you?

glensc:
the tn/3 url corresponds to: https://datatables.net/manual/tech-notes/3

glensc:
ok, perhaps the init twice becames that it's first initialized via css class:
[templates/manage/users_list.tpl.html:100](https://github.com/eventum/eventum/blob/v3.1.3/templates/manage/users_list.tpl.html#L100)


and then via javascript method:
[templates/manage/users_list.tpl.html:65](https://github.com/eventum/eventum/blob/v3.1.3/templates/manage/users_list.tpl.html#L65)

but i'm still unable to reproduce in my test env

glensc:
nevermind, line 100 just defines the class line 65 activates. i don't see how it could be activated twice

glensc:
wow, in that broken instance:
```
  $(function() { 
    console.log('init');
  });
```

prints:
```
init
init
```

how come `jQuery.ready()` is invoked twice?!

this also explains why pasting images from clipboard uploaded them twice.

glensc:
fucking Smarty. it can fucking burn in hell. it's `{block ... append}` is total bullshit.

adding this to `templates/extra_header.tpl`
```
{block "javascripts" append}
  <script type="text/javascript">
  </script>
{/block}
```

random other `{block "javascripts" append}` filled twice to resulting html.

first it "appends" items in included templates in wrong order. now this. !@#$!@%#$% :pig:

glensc:
created reproducer gist: https://gist.github.com/glensc/59d329718118dc4ef6bedacbe22406a8

showing both bugs:

given the `test.tpl`:
```smarty
{extends "manage.tpl"}
{block "javascripts" append}
{/block}
{include file="form.tpl"}
{include file="list.tpl"}
```

# includes in wrong order
```
$ git clone https://gist.github.com/59d329718118dc4ef6bedacbe22406a8.git -b test-2 test-2
$ cd test-2
$ composer install
[...]

PR #264, opened 2017-05-12
add userfile.js, userscript.css support

glensc:
Adds userscript.js, userstyle.css support.
Inspired by DokuWiki implementation.
- https://www.dokuwiki.org/devel:javascript
- https://www.dokuwiki.org/devel:css#user_styles

To add custom JavaScript or custom CSS to every Eventum page,
Place "userscript.js", "userstyle.css" to "config/" dir
and they would be automatically included to every page.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -17,6 +17,7 @@ Consult [Upgrade] wiki page how to upgrade from previous versions.
 - use ctrl/cmd enter to submit forms (@glensc, #255)
 - quote custom field names (@glensc, #258)
 - drop Mail_rfc822 PEAR Mail requirement (@glensc, #256)
+- add userfile.js, userscript.css support (@glensc, #264)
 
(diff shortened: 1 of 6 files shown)


glensc:
using such specific js/css inject solution, i can solve #192, because don't need to use `extra_header.tpl.html` anymore.

thus avoid that smarty problem: https://github.com/smarty-php/smarty/issues/303

balsdorf:
I wonder if we should have them in a subdirectory (static/ maybe?) but don't have a good name for it or a strong opinion.

This is a nice solution and cleaner than the current process. I'm good with merging it.

phavel:
Please, add info to wiki. Thanks.

glensc:
Wiki is open for edits for any contributor.

phavel:
@glensc  My english is poor, I do not dare write in official documents.

PR #270, opened 2017-05-20
drop old db migrate system

glensc:
The old db migrate system written on our own has served us well almost 9 years (first commit: e821e32e).

It has made upgrade simple one command run since Eventum 2.2 (2009-01-14) to 3.2.0 (2017-05-20).
it has been replaced by phinx: #235, so we don't have to maintain db migrator on our own.

upgrading 3.2.x onwards series requires that you update to 3.2.0 first.

should write that to wiki:
https://github.com/eventum/eventum/wiki/Upgrading

but that page is written in awkward manner with "upgrading from", but i need to write "upgrade to", @balsdorf help.


balsdorf:
@glensc I tried to clarify the docs page, not sure if it helped.

Other then that, I'm good to go on this. A bit of a pain for people who don't keep up with releases, but I don't see any way around it.

glensc:
we can merge it later 3.2.x releases. but does not really make any difference, does it

balsdorf:
So unless I'm missing something, we shouldn't have merged this yet.

I just tried to upgrade one of my 3.1.10 installations and it starts from the first phinx migration because `upgrade/patches/67_phinxlog.sql` was never run (because it didn't exist in 3.1.10)

glensc:
why are you upgrading from 3.1.10, if the docs clearly say you need to upgrade to 3.2.0 **first**.

balsdorf:
3.2.0 is still marked as pre-release (
https://github.com/eventum/eventum/releases/tag/v3.2.0) so I mistakenly
thought all unreleased changes were part of it. I see now that it is tagged
before the migrations were removed.

glensc:
what you mean "still"? same version can't evolve over time as stable. new release needs to be made.

3.2.0 is very experimental as lots of possibly breaking changes in 3.2.x, so i marked it as "pre-release".

glensc:
besides, i did inform you about plan to create 3.2.0 tag this way in gitter: https://gitter.im/eventum/eventum?at=591c013c2b926f8a67633882

vladsf:
Hello!

I am happily running Eventum 2.3.6 :) What would be the right way to upgrade? 
2.3.6 -> 3.0.0-pre1 -> 3.2.0 (or 3.1?) -> 3.4.latest?

glensc:
See wiki: https://github.com/eventum/eventum/wiki/Upgrading

PR #272, opened 2017-05-20
add Event/EventDispatcher to Eventum

glensc:
Adds Events to Eventum using [Symfony Event Dispatcher](https://symfony.com/doc/current/components/event_dispatcher.html).

First implementation is handling projectless crypto enable/disable events (#165). I am probably the only one using those workflow methods :)

Example Extension updated with the sample that encrypts/disables `$config['ftp']['password']` field when encryption is turned on/off from administration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - add `$min_role` to base controller (@glensc, #261)
 - require `iconv` and `mbstring` extensions (@glensc, #269)
+- add Event/EventDispatcher to Eventum (@glensc, #272)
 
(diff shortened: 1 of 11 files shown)


glensc:
i wonder perhaps need to rename `CryptoSubscriber`  to `CryptoConfigSubscriber` or `ConfigCryptoSubscriber` as it deals with config encrypt/decrypt.

and as the documentation suggests, perhaps event names need to be renamed as well to indicate when they are ran.

glensc:
and should prefix `WorkflowEvents` event names with `workflow.`

PR #277, opened 2017-07-12
Add "Status Change Date" field to issue.

balsdorf:
I've added a new field to the issue to track when the status was last changed and I've added it to the list issues page as well.

This name conflicts with some old, confusing functionality [here](https://github.com/eventum/eventum/blob/e4c32c4579673a766e7fffac23c66ee103e62965/lib/eventum/class.issue.php#L2401) (linked from the admin as "Customize Issue Listing Screen") which allows you to choose your own date field to display based on what the status is. I can see some value in this functionality, but I don't use it and am not sure if anyone else does either.

I would like to either rename it to "Status Action Date" (or something better, I'm open to suggestions) or remove the functionality.

Thoughts?


glensc:
new field seems ok, not breaking anything. column "Status Change Date" is perfect name IMHO.

as for current functionality you mean "Last Action Date"?

balsdorf:
Not "Last Action Date",that field is fine. The old field I want to rename / delete is called "Status Change Date" (which is why I titled the new column "Last Status Change Date" until we . See [here](https://github.com/eventum/eventum/blob/e4c32c4579673a766e7fffac23c66ee103e62965/lib/eventum/class.display_column.php#L204) and [here](https://github.com/eventum/eventum/blob/e4c32c4579673a766e7fffac23c66ee103e62965/src/Controller/Manage/CustomizeListingController.php)

slay123:
Hello. I couldn't see that feature in our system. So perhaps we don't even use it?

What must definitely stay is "Submitted date", "Last updated date" and our custom field "Deadline" (which is actually rarely used). These are date selector fields that we use.

balsdorf:
@slay123 Those fields will stay for sure. Like I said, I don't think anyone really uses the old column, but in case they do I'm happy to rename the old field to "Status Action Date"

glensc:
@balsdorf as @slay123 concern is preserved, do whatever you please with the old field :) it's hard to give opinion on something that i've never seen :)

balsdorf:
Alright I've renamed that old field, a lot more places to change once I decided to be thorough :)

Let me know if you see any problems

PR #278, opened 2017-07-25
emit history.add events when history entry is added

glensc:
just an experiment, is it useful yet, don't know :)

to test, add to `config/setup.php`:
```php

  'extensions' => [
    'Eventum\\Extension\\ExampleExtension' => dirname(__DIR__) . '/docs/examples/extension/ExampleExtension.php',
  ],
```

real usecase could be removing irc related code from eventum-core and move it into extension.
or some kind of activity monitor.


glensc:
my current irc notification is handled by these workflow methods:
- handleNewEmail
- handleIssueClosed
- handleBlockedEmail
- handleAssignmentChange
- handleScmCommit

glensc:
merging, can drop/improve later.

PR #279, opened 2017-07-27
use mnapoli/silly for cli applications

glensc:
We have Symfony\Console as dependency since introduction of Phinx.

For long time wanted to convert the console scripts using Symfony\Console, they were even moved to classes (#137) with having in mind porting to Symfony\Console one day.

So today found [Silly CLI micro-framework based on Symfony Console.](http://mnapoli.fr/silly/) and couldn't resist.

Symfony\Console benefits:
- unified commandline arguments passing (`--verbose`, `--quiet`)
- colored output!

probably could easily add debug messages with verbose increasing. currently no script is able to print more information if wanted to debug script execution.


glensc:
Due the way Symfony\Console works, scripts that use non-option arguments, their invocation has been changed. Commands that used only options or none, for them it was possible to setup default command to be run without arguments. For these invocation has not changed.

| Before  | After |
| ------------- | ------------- |
| bin/check_reminders.php [--debug] | bin/check_reminders.php [--debug]  |
| bin/download_emails.php [username] [hostname] [mailbox] | bin/download_emails.php mail:download [username] [hostname] [mailbox] |
| bin/monitor.php [-q] | bin/monitor.php [-q\|--quiet] |
| bin/process_all_emails.php [filename] | bin/process_all_emails.php mail:route [filename] |
| bin/process_mail_queue.php | bin/process_mail_queue.php |
| bin/truncate_mail_queue.php | bin/truncate_mail_queue.php [-q\|--quiet] [--interval=] |

glensc:
perhaps should just have single entrypoint for those cli commands (except bin/upgrade.php and bin/irc-bot.php because they are different) and deprecate current ones.

what the name would be?

- **bin/cli.php**
- **bin/eventum.php**
- **bin/console.php**
- ...your idea here?

glensc:
looks like with symfony/console 3.2 it's possible to enable $isSingleCommand:

http://symfony.com/doc/3.2/components/console/single_command_tool.html

and given that example don't even need silly library much (ok, it does provide DI and simple interface to define all arguments and options):

```php
#!/usr/bin/env php
<?php
require __DIR__.'/vendor/autoload.php';

use Symfony\Component\Console\Application;
use Symfony\Component\Console\Input\InputArgument;
use Symfony\Component\Console\Input\InputInterface;
use Symfony\Component\Console\Input\InputOption;
use Symfony\Component\Console\Output\OutputInterface;

(new Application('echo', '1.0.0'))
  ->register('echo')
      ->addArgument('foo', InputArgument::OPTIONAL, 'The directory')
      ->addOption('bar', null, InputOption::VALUE_REQUIRED)
      ->setCode(function(InputInterface $input, OutputInterface $output) {
          // output arguments and options
      })
  ->getApplication()
  ->setDefaultCommand('echo', true) // Single command application
  ->run();
```

balsdorf:
Sorry I didn't get a chance to review this right away but I think it is all good. Our cli scripts have been rather painful.

I like the idea of a single entrypoint as well for a future improvement, `bin/eventum.php` would be my first choice.

cpinfold:
My suggestion is:

bin/process_jobs.php

I also like the idea of consolidating.. would be even better if we could
process multiple commands from one cron job.

Eg. bin/process_jobs.php mail:download ... mail:send reminders:process etc..
bin/process_jobs.php mail:downloadall (no arguments needed, processes all
active accounts)

Craig.

glensc:
dislikes:
- `bin/eventum.php` we kind of already have eventum cli took named `eventum.phar`, could cause confusion
- `bin/process_jobs.php` - not all commands "process jobs", `bin/monitor.php` for example currently is for checking system health, `bin/extension.php` is for managing eventum extensions

processing multiple commands, i had this idea too, but mostly because i invoke same command repeatedly for different projects in serial manner:

```
bin/download_emails.php eventum@example.org example.org INBOX.support
bin/download_emails.php eventum@example.org example.org INBOX
```

balsdorf:
Good point on `bin/eventum.php`. Django uses `manage.py` but I can see how that wouldn't always be appropriate. I'm fine with `console.php`, `command.php` or `manage.php` but I don't think any name is going to be 100% accurate unfortunately.

@cpinfold You actually can run multiple commands in one cronjob, just separate the commands with a ';'. 
```
* * * * * bin/download_emails.php eventum@example.org example.org INBOX.support; bin/download_emails.php eventum@example.org example.org INBOX
```

cpinfold:
Thanks Bryan.. hadn't thought of doing it that way.

+1 for command.php

It would also be nice to call:

bin/command.php download:all
or bin/command.php mail:downloadall

And let it go through all the accounts.  I have a few accounts checked less
frequently, but wouldn't if I could call one command.

Craig.

PR #280, opened 2017-07-28
rpc: setIssueStatus: validate parameters server side

glensc:
this is how it should had always been done.

needed by https://github.com/eventum/cli/pull/3

i replaced whatever access checks (`checkIssuePermissions`, `checkIssueAssignment`) there was with `Issue::canUpdate`. this should be sufficient right @balsdorf ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Upgrading to 3.2.x versions requires that you upgrade to 3.2.0 version first.
 - Fix bug auto switching project on update issues page (@balsdorf, 6ffecfc)
 - Catch exception if invalid message is added to mail queue (@balsdorf, 0e55ae2)
 - add real 'Status Change Date' column and rename old column to 'Status Action Date' (@balsdorf, #277)
+- rpc: setIssueStatus: validate parameters server side (@glensc, #280)
 
 [3.2.2]: https://github.com/eventum/eventum/compare/v3.2.1...master
 
--- a/lib/eventum/class.command_line.php
+++ b/lib/eventum/class.command_line.php
@@ -440,31 +440,6 @@ class Command_Line
      */
     public static function setIssueStatus(&$client, $auth, $issue_id, $new_status)
     {
-        $details = self::checkIssuePermissions($client, $auth, $issue_id);
-        self::checkIssueAssignment($client, $auth, $issue_id);
-
-        // check if the issue already is set to the new status
-        if ((strtolower($details['sta_title']) == strtolower($new_status)) ||
-                (strtolower($details['sta_abbreviation']) == strtolower($new_status))) {
-            self::quit("Issue #$issue_id is already set to status '" . $details['sta_title'] . "'");
-        }
-
-        // check if the given status is a valid option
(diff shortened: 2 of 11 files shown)


glensc:
> Finally, I'd move the logic from Issue::setRemoteStatus into the RemoteAPI class and remove that method.

yes. i thought that, but as another method uses that call, planned such change for the future

glensc:
ok, i'll add `Access::canChangeStatus` check as well, and if remote calls is not enabled it should throw error anyway?

also i've deprecated `Issue::canUpdate` to use `Access::canUpdateIssue` directly. kept it around because maybe some workflow method uses it.

glensc:
do you want to include check `RemoteApi::mayChangeIssue` (used by `Command_Line::checkIssueAssignment()`) into the `setIssueStatus`? or it should be rather added to `Access::canChangeStatus`?

balsdorf:
`RemoteApi::mayChangeIssue` is poorly named right now since it just checks if the issue is assigned to the current user. It should be renamed and added to `setIssueStatus` though doing so will complicate things.

If you can find a good way to add the check and confirmation prompt to `setIssueStatus` I would, otherwise I'd leave the separate call to `Command_Line::checkIssueAssignment()` in. It unfortunately adds an extra round trip but it might be preferable to cluttering up each individual method.

glensc:
@balsdorf ported the functions from Command_Line class. this is what you had in mind?

it involves duplication, and likely the methods could be improved to fetch data directly, but i do not wish to do those optimizations right now.

glensc:
oh, i did not remembered i modified old cli app. but nevertheless the changes are moved to server side. the prompting for invalid input may be broken, is it worth of it to fix?

balsdorf:
The cleanup of the code is good, but the assignment check has lost functionality. Now you cannot continue past the warning.
```
# on master
bryan@odin:cli$ ./eventum.php 1 ss discovery
WARNING: You are not currently assigned to issue #1.
These are the current details for issue #1:
         Summary: sssssssss
          Status: requirements
      Assignment: Admiñ User
  Auth. Repliers: Bryan D Alsdorf
Are you sure you want to change this issue? [default: y] -> y
OK - Status successfully changed to 'discovery' on issue #1
bryan@odin:cli$ git checkout glensc-rpc-set-status-command 
Switched to branch 'glensc-rpc-set-status-command'
bryan@odin:cli$ ./eventum.php 1 ss requirements
ERROR: You are not currently assigned to issue #1.
```
Now on the current master there actually is a bug that prevented this confirmation from showing up, looks like ever since we switched to autoloading classes. [This variable](https://github.com/eventum/eventum/blob/e9911aa619daccd0dd3ed6ef958e88ad0a2df9d1/lib/eventum/class.command_line.php#L16) doesn't end up on in the global scope so [this](https://github.com/eventum/eventum/blob/e9911aa619daccd0dd3ed6ef958e88ad0a2df9d1/lib/eventum/class.command_line.php#L323) check fails.

glensc:
as i understand that reverting changes in command_line class restores behavior you are interested in?

balsdorf:
Yes, with the caveat from above that autoloading broke it due to scope issues.

glensc:
@balsdorf reverted

PR #289, opened 2017-09-08
add events for user creation/update

glensc:
this would allow workflow project to fill external_id, or similar changes.


glensc:
@balsdorf why such restriction? why can't the field be nullable?

```
( ! ) Fatal error: Uncaught exception 'PDOException' with message
'SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'usr_external_id' cannot be null'```

balsdorf:
@glensc Yeah that should definitely be nullable. No idea why it isn't currently.

PR #291, opened 2017-09-09
WIP: add experimental markdown rendering

glensc:
this works surprisingly well. currently enabled in issue description area.

to test this (besides getting code from this branch and updating composer):

define in config/setup.php
```php
return [
    'markdown' => 'enable',
...
```


inguin:
Works great for me. Only issue is that clicking the 'fixed width' link messes up the display. This can be easily fixed by replacing the description with `$issue.iss_original_description` when switching to fixed with display.

glensc:
can you propose patch?

inguin:
Maybe like this: inguin@24f6510f32891f1ae30c85b1b1833add338581e2

glensc:
checked in as dc972f878...4ddf69fc3

rtmaki:
It's one solution, but not exactly what I was hoping for. I was hoping for the ability to copy directly from a web page, or just take a screen shot and paste in the image from the buffer etc. This is something that can be done with WSYWIGs sometimes. Even if we could just upload the image, and then have an insert image icon so that we can have a quick way to format images properties (height, width, src= etc.) In the meantime however, your solution will work for images, just a bit more clunky for the non technical users. Thanks for your update! :-)

glensc:
markdown should support html to some extent, but i haven't tested this.

rtmaki:
With turning on Markdown, and then installing CKEditor on any textarea I wanted, it works amazingly well now. With CKEditor with image upload support, I can copy and paste from the buffer, and can also drag and drop files etc. This would be a cool feature to add out of the box :-) .. thanks for your help! This can be closed now!

glensc:
so, submit your changes as pull request!

PR #295, opened 2017-09-20
unify lengths of header fields

### LINKED ISSUE #266: length of email headers
i was investigating why code from here:
[lib/eventum/class.support.php:1551-1552](https://github.com/eventum/eventum/blob/v3.1.10/lib/eventum/class.support.php#L1551-L1552)

was throwing mail parse errors:
```
Zend\Mail\Exception\InvalidArgumentException(code: 0):
'blah, blah <blah.blah' can not be matched against dot-atom format at ...
```

seems that `to_cc` was truncated to 255 chars at the time data was inserted.
i checked the database, and the field is longer now, however can't find patch which increased the size, and seems they are inconsistent:

```mysql
  `sup_from` varchar(255) NOT NULL DEFAULT '',  
  `sup_to` text,
  `sup_cc` mediumtext,
  `sup_subject` varchar(255) NOT NULL DEFAULT '',
```

we should decide what the field type and length are and unify it everywhere, possibly fixing corrupted fields as well.

**PS**: in pristine installation `sup_cc` is:
```mysql
  `sup_cc` text,
```

glensc:
resolves #266

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Upgrading to 3.3.0 versions requires that you upgrade to 3.2.0 version first.
 - Drop support for old setup.php formats (@glensc, #288)
 - Add script to manage extensions from CLI (@glensc, #290)
 - Use mnapoli/silly for cli applications (@glensc, #279)
+- Unify lengths of header fields (@glensc, #266, #295)
 
 [3.3.0]: https://github.com/eventum/eventum/compare/v3.2.3...master
 
--- a/db/migrations/20170920202640_eventum_sup_header_length.php
+++ b/db/migrations/20170920202640_eventum_sup_header_length.php
@@ -1,0 +1,33 @@
+<?php
+
+/*
+ * This file is part of the Eventum (Issue Tracking System) package.
+ *
+ * @copyright (c) Eventum Team
+ * @license GNU General Public License, version 2 or later (GPL-2+)
+ *
+ * For the full copyright and license information,
+ * please see the COPYING and AUTHORS files
+ * that were distributed with this source code.
+ */
+
+use Eventum\Db\AbstractMigration;
+
+class EventumSupHeaderLength extends AbstractMigration
+{
+    const HEADER_LENGTH = 4096;
+
+    /**
+     * here's no down() because we only changed column width
+     */
+    public function up()
+    {
+        $type = self::PHINX_TYPE_STRING;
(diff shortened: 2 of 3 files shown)


glensc:
note: this is not big table. even my system with ~120k rows:

```
$ LC_ALL=C ls -lh support_email.*
-rw-rw---- 1 mysql mysql 23M Sep 20 19:54 support_email.MYD
-rw-rw---- 1 mysql mysql 12M Sep 20 19:54 support_email.MYI
-rw-rw---- 1 mysql mysql 13K Mar 30  2016 support_email.frm
```

glensc:
@balsdorf is there difference (what kind) of `NOT NULL DEFAULT ''` and  `NOT NULL`?

```diff
@@ -39,10 +39,10 @@
   `sup_customer_id` varchar(128) DEFAULT NULL,
   `sup_message_id` varchar(255) NOT NULL DEFAULT '',
   `sup_date` datetime NOT NULL DEFAULT '0000-00-00 00:00:00',
-  `sup_from` varchar(4096) NOT NULL,
-  `sup_to` varchar(4096) NOT NULL,
-  `sup_cc` varchar(4096) NOT NULL,
-  `sup_subject` varchar(4096) NOT NULL,
+  `sup_from` varchar(4096) NOT NULL DEFAULT '',
+  `sup_to` varchar(4096) NOT NULL DEFAULT '',
+  `sup_cc` varchar(4096) NOT NULL DEFAULT '',
+  `sup_subject` varchar(4096) NOT NULL DEFAULT '',
   `sup_has_attachment` tinyint(1) NOT NULL DEFAULT '0',
   `sup_removed` tinyint(1) NOT NULL DEFAULT '0',
   PRIMARY KEY (`sup_id`),
```

balsdorf:
From: https://dev.mysql.com/doc/refman/5.7/en/data-type-defaults.html

> For data entry into a NOT NULL column that has no explicit DEFAULT clause, if an INSERT or REPLACE statement includes no value for the column, or an UPDATE statement sets the column to NULL, MySQL handles the column according to the SQL mode in effect at the time:

> If strict SQL mode is enabled, an error occurs for transactional tables and the statement is rolled back. For nontransactional tables, an error occurs, but if this happens for the second or subsequent row of a multiple-row statement, the preceding rows will have been inserted.

> If strict mode is not enabled, MySQL sets the column to the implicit default value for the column data type.

We don't run in strict mode by default so really they behave the same. In this case, `NOT NULL DEFAULT ''` is probably the safest option since I'm not sure if the code handles NULL values properly

Issue #300, opened 2017-10-02
3.3.0 release


glensc:
- [x] update min php requirement to 5.6.0 (@balsdorf)
- [x] some docs how to upgrade (does one need extra disk space or not, or how to estimate needed disk space). #254

balsdorf:
Disk space question is a good one. I'll do some tests and let you know.

glensc:
there are two kind of disk requirements:
- uprading 3.2.3 -> 3.3.0
- migrating from one storage type to another

in this issue scope, only the first one is interesting

balsdorf:
For upgrading from 3.2.3 -> 3.3.0 there should be no additional disk space required. The upgrade process just adds the new tables and columns but leaves existing attachments in the current table. New attachments will be inserted into the new table but unless you run `migrate_storage_adapter.php` it won't move anything.

glensc:
thanks for the info! i'll copy that to original PR too. should we add something to changelog.md too or as nothing special, not necessary?

glensc:
i think i'll roll out 3.3.0 release, today-tomorrow, unless @balsdorf want to do that yourself.

balsdorf:
Nothing is needed in changelog.md regarding disk space, but not sure if we want to add text letting people know they can switch to a different adapter? I'm undecided as I view this as an advanced feature that most people won't need so I don't know if we want to advertise it. I went ahead and added a [new page in the wiki](https://github.com/eventum/eventum/wiki/Attachments) regarding it. Think this is good enough or we should advertise it in the changelog?

And please go ahead with the release when you can.

phavel:
On the contrary, it is a important function.
I'm waiting for it and I am happy it will be released. My MySQL storage on a web host is already full because of attachments.

Thanks for the function!

balsdorf:
@phavel I'm not worried about people like you who are reading these PRs! :) Once you try it out let me know any feedback you have, updated docs are appreciated as well.

glensc:
@balsdorf damn, your analyze is wrong! EventumAttachments does alter on issue_attachment and issue_attachment_file tables! meaning there's need for double of those tables size!

glensc:
@balsdorf how you propose to solve this, add the new column of `issue_attachment_file` table to some new table perhaps instead?

balsdorf:
What version of MySQL are you using and what engine?

For InnoDB newer than 5.6 it table adds / drops are handled INPLACE. For
[...]

PR #301, opened 2017-10-04
pdo: handle PDOException HY000 non-numeric values

glensc:


glensc:
found this bug with major disaster when tried to do test upgrade to 3.3.0 using snapshot build, and disk got full (see [here](https://github.com/eventum/eventum/issues/300#issuecomment-334227753))

killed running query because unable to free enough disk space. and top of all that mysqld itself crashed too. and after that some queries started to spew errors like:

```

app.ALERT: Fatal Error (E_ERROR): Wrong parameters for Exception([string $exception [, long $code [, Exception $previous = NULL]]]) {"code":1,"message":"Wrong parameters for Exception([string $exception [, long $code [, Exception $previous = NULL]]])","file":"/usr/share/eventum/src/Db/Adapter/PdoAdapter.php","line":189} {"file":null,"line":null,"class":null,"function":null,"memory_usage":"3.75 MB","memory_peak_usage":"3.75 MB","version":"3.2.2","usr_id":""}
```

appears php creates [PDOxception](http://php.net/manual/en/class.pdoexception.php) class whose `$code` data type is `string`, and even it extends [RuntimeException](http://php.net/manual/en/class.runtimeexception.php) whose `$code` data type is `int`, it's impossible to construct such exception object from userspace code. used reflection for that when testing such scenario.

/cc @balsdorf.

glensc:
seems exceptions now properly converted. however mail_queue should be fixed not to hide exceptions. but don't want to make changes now as that would conflict with #263.

PR #315, opened 2017-10-10
auth/ldap: add commandline script

glensc:


glensc:
here's interesting problem i encountered.

appears Zend\Config approach of merging array's key integer values means append whatever is merged into: [src/Config.php:334-335](https://github.com/zendframework/zend-config/blob/release-2.4.13/src/Config.php#L334-L335)

this resulted each time when ldap preferences were saved `default_role` to grow:

```php
  'default_role' =>                                                                                                      
  array (                                                                                                                
    1 => '2',                                                                                                            
    2 => '2',                                                                                                            
    3 => '2',                                                                                                            
    4 => '2',                                                                                                            
    5 => '2',                                                                                                            
    6 => '2',                                                                                                            
  ),
```

[src/Controller/Manage/LdapController.php:79](https://github.com/eventum/eventum/blob/v3.3.1/src/Controller/Manage/LdapController.php#L79)

even if data what was actually saved was just two entries.

so i came up with this method, to have string type of array keys. unfortunately that only did not solve the problem, so the method not committed, but saving here for future times.

```php

    /**
     * ensure keys are type string,
     * as Zend\Config would otherwise append values when merging
     *
     * @param array $array
     * @return array
     * @link https://stackoverflow.com/a/35180513/2314626
     */
    private function stringifyKeys($array)
    {
        $obj = (object)[];
        foreach ($array as $key => $value) {
            $obj->{(string)$key} = $value;
        }

        return (array)$obj;
    }
```

PR #316, opened 2017-10-13
Prevent users from viewing "removed" notes

balsdorf:
This prevents users from accessing notes marked as removed as well as documenting why we use `not_removed` instead of actually deleting them.

@glensc , do you think it would be better to actually delete the note if it wasn't routed from an email (i.e. `not_message_id` is blank) ? Another option would be for removed notes set the title and body to be null.

While I was updating code I did some general cleanup such as stop hiding errors and removing code flagged in FIXMEs (once I confirmed it was no longer needed)

### DIFF
--- a/lib/eventum/class.note.php
+++ b/lib/eventum/class.note.php
@@ -63,7 +63,7 @@ class Note
      * Retrieves the details about a given note.
      *
      * @param   int $note_id The note ID
-     * @return  array The note details
+     * @return  false|array The note details
      */
     public static function getDetails($note_id)
     {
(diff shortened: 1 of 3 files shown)


lib/eventum/class.note.php
```diff
@@ -63,7 +63,7 @@ public static function getSideLinks($issue_id, $not_id)
      * Retrieves the details about a given note.
      *
      * @param   int $note_id The note ID
-     * @return  array The note details
+     * @return  false|array The note details
```
glensc: `bool` not `false`

glensc:
merging; the actual delete can be made in separate MR

PR #323, opened 2017-11-09
Support multiple selections in search fields

inguin:
* Convert the search fields in the quick filter toolbar and advanced
  search to multi-select fields
* For saved searches, store the selection as string of comma-separated
  values
* Add DB migration to change column types in `custom_filter` to string
* List of changed fields:
  - Assigned (cst_users)
  - Category (cst_iss_prc_id)
  - Priority (cst_iss_pri_id)
  - Product  (cst_pro_id)
  - Release  (cst_iss_pre_id)
  - Reporter (cst_reporter)
  - Severity (cst_iss_sev_id)
  - Status   (cst_iss_sta_id)

### DIFF
--- a/db/migrations/20171109145542_eventum_custom_filter_multiselect.php
+++ b/db/migrations/20171109145542_eventum_custom_filter_multiselect.php
@@ -1,0 +1,49 @@
+<?php
+
+/*
+ * This file is part of the Eventum (Issue Tracking System) package.
+ *
+ * @copyright (c) Eventum Team
+ * @license GNU General Public License, version 2 or later (GPL-2+)
+ *
+ * For the full copyright and license information,
+ * please see the COPYING and AUTHORS files
+ * that were distributed with this source code.
+ */
+
+use Eventum\Db\AbstractMigration;
+
+class EventumCustomFilterMultiselect extends AbstractMigration
+{
+    const COLUMNS = [
+        'cst_iss_pri_id' => 'cst_priorities',
+        'cst_iss_sev_id' => 'cst_severities',
+        'cst_reporter'   => 'cst_reporters',
+        'cst_iss_prc_id' => 'cst_categories',
+        'cst_iss_sta_id' => 'cst_statuses',
+        'cst_iss_pre_id' => 'cst_releases',
+        'cst_pro_id'     => 'cst_products',
+    ];
+
+    public function up()
+    {
+        $table = $this->table('custom_filter');
+        foreach (self::COLUMNS as $old_name => $new_name) {
+            $table->changeColumn($old_name, 'string', ['length' => 255, 'null' => true]);
+            $table->renameColumn($old_name, $new_name);
+        }
+        $table->changeColumn('cst_users', 'string', ['length' => 255, 'null' => true]);
+        $table->save();
+    }
+
+    public function down()
+    {
+        $table = $this->table('custom_filter');
+        foreach (self::COLUMNS as $old_name => $new_name) {
+            $table->changeColumn($new_name, 'integer', ['length' => 10, 'null' => true, 'signed' => false]);
+            $table->renameColumn($new_name, $old_name);
+        }
+        $table->changeColumn('cst_users', 'string', ['length' => 64, 'null' => true]);
+        $table->save();
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 Upgrading to 3.4.x versions requires that you upgrade to 3.2.0 version first.
 
 - Use Doctrine as ORM for SCM and IssueAssociation (@glensc, #307)
+- Support multiple selections in search fields (@inguin, #323)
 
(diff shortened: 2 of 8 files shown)


glensc:
don't like the idea of having `_id` column of type string. @balsdorf what you think?

inguin:
I agree with @glensc, a column named 'cst_iss_sta_id' might be expected to directly reference 'iss_sta_id' in the issue table.

I tried to keep the schema changes to a minimum, but it should't be a big deal to rename the columns if you'd prefer that.

balsdorf:
I'll do a deeper dive into the code later but like @glensc I agree that the column names should be changed.   I'd be okay with cst_categories, cst_severities, etc

inguin:
I renamed the `custom_filter` columns according to your recommendations. I also split the feature into two commits, one for the DB migration and one for the actual multi-select search.

db/migrations/20171109145542_eventum_custom_filter_multiselect.php
```diff
@@ -0,0 +1,47 @@
...
+use Eventum\Db\AbstractMigration;
+
+class EventumCustomFilterMultiselect extends AbstractMigration
+{
+    const COLUMNS = [
+        'cst_iss_pri_id' => 'cst_priorities',
+        'cst_iss_sev_id' => 'cst_severities',
+        'cst_reporter'   => 'cst_reporters',
+        'cst_iss_prc_id' => 'cst_categories',
+        'cst_iss_sta_id' => 'cst_status',
```
glensc: should be `cst_statuses` then if it contains multiple values like other fields.

db/migrations/20171109145542_eventum_custom_filter_multiselect.php
```diff
@@ -0,0 +1,47 @@
...
+        'cst_iss_pre_id' => 'cst_releases',
+        'cst_pro_id'     => 'cst_products',
+    ];
+
+    public function up()
+    {
+        $table = $this->table('custom_filter');
+        foreach (self::COLUMNS as $old_name => $new_name) {
+            $table->changeColumn($old_name, 'string', ['length' => 64, 'null' => true]);
```
glensc: i would use `255` here as there's no special reason to have `64`

db/migrations/20171109145542_eventum_custom_filter_multiselect.php
inguin: I always have trouble making up my mind about that word. The grammatically correct plural is 'status', but it sounds weird and might conflict with a singular use. I'll change it.

db/migrations/20171109145542_eventum_custom_filter_multiselect.php
inguin: ACK

inguin:
I renamed  `cst_status` to `cst_statuses` and changed the length of string fields from 64 to 255. I also added a migration step for `cst_users` which for some reason already was a varchar(64) before.

glensc:
tried the branch locally (4ea804435), but get dozen of warning notices.

db patch applied
```
 == 20171109145542 EventumCustomFilterMultiselect: migrating
```

```
( ! ) Warning: Invalid argument supplied for foreach() in /home/glen/scm/eventum/eventum/lib/eventum/class.search.php on line 502
--


1 | 0.0003 | 253800 | {main}( ) | .../list.php:0
2 | 0.0777 | 5674544 | Eventum\Controller\BaseController->run( ) | .../list.php:17
3 | 0.0949 | 6969528 | Eventum\Controller\ListController->prepareTemplate( ) | .../BaseController.php:78
4 | 0.1072 | 7210520 | Search::getListing( ) | .../ListController.php:206
5 | 0.1153 | 7868960 | Search::buildWhereClause( ) | .../class.search.php:336


```

inguin:
Added a commit to fix all warnings. My modified `Search::getParam()` wasn't very good at assuring the return values would be arrays; default values or values from the user's search profile could still be returned as scalars.

I'll squash the fixes into the actual implementation commit once the pull request is approved.

glensc:
checked out the branch locally.

i don't like that the ui change makes search filters taking 50% of the screen.

such large options block belongs to advanced search imho. could the default view be still single field based but allow multiselects on advanced search? and advocate using saved searches for common filters?


before:
image

after:
image

inguin:
I understand your concern (although you'd have to have a pretty low screen resolution for the search bar to fill half of the height). We've been using the tall filter fields without complains for at least a decade, but that doesn't mean other long-time Eventum users will be happy with the change. I could reduce the number of rows in the select boxes, but that makes then less convenient to use.

Limiting the multi-selection to the Advanced Search page makes sense and would be very easy to implement. The only issue is that the user won't be able to repeat or refine the search afterwards. Clicking the Search button will lose part of his selection.

A possible solution for this would be to introduce a special 'multiple' option that is displayed and selected for search fields with more than one option selected.

inguin:
I added an experimental commit reverting the quick filter back back to plain select boxes and introducing a special "multiple selections" item.

glensc:
looks good now, i think it's ok to be merged

PR #324, opened 2017-11-09
Don't show inactive users in notification list

balsdorf:
Users with a status of 'inactive' already do not get emails due to the mail queue checking their status before sending. This PR cleans hides them from the NL when viewing an issue. They are still visible if you go to edit the NL.

I've also cleaned up [Notification::getUsersByIssue](https://github.com/eventum/eventum/blob/7ca21595fdaed4f2cbb58b72f3919b5f81f34d72/lib/eventum/class.notification.php#L543) for notes to only return users with a role of Standard User or above. This restriction is already enforced elsewhere in the code but returning all users here is incorrect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Eventum 3.3.x requires PHP 5.6.
 - Fix sending same reminder repeatedly (@phavel, #321)
 - Use symfony/ldap for ldap connection (@glensc, #322)
 - Make `git diff`, `git log` on `*.po`/`*.pot` files less useful (@glensc, #325)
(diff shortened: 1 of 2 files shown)


glensc:
btw, inactive users get show in banner "email sent to" when you update issue or post note.

See [$_EVENTUM_LAST_NOTIFIED_LIST](https://github.com/eventum/eventum/blob/7ca21595fdaed4f2cbb58b72f3919b5f81f34d72/lib/eventum/class.notification.php#L971)

does this solve that problem too?

balsdorf:
Yes this solves that problem as well

PR #326, opened 2017-11-13
Fix mail submission with PHP7

inguin:
Mail_Helper::$headers is declared as a string but later used as an
array. PHP 5 would silently change its type to array; PHP 7 raises an
'Illegal string offset' warning instead.


glensc:
this code is dropped in #263, but too big changes to take risk and merge it. should test it a bit in my own production first.

glensc:
looks like i pressed "merge" while intended to press "comment". oh well :D

PR #329, opened 2017-11-27
LDAP improvements

inguin:
- Make user ID attribute configurable to support integration with Microsoft AD
- Fix exception when auto-creating user accounts


glensc:
please add default in [ldap_auth_backend](https://github.com/eventum/eventum/blob/ed9707aebc82a9ee073c43f02dfc29baa44a6d8d/lib/eventum/auth/class.ldap_auth_backend.php#L475)

glensc:
cool. another ldap user.

inguin:
Added the following:
- Default value as requested
- Changelog entry
- Help page for LDAP configuration

Touching the user filter won't do any good: the current default won't work with AD anyway because the principal names contain an '@' character (e.g. `user@domain.company.com`) and would be treated as mail addresses. Integrating with AD will currently require specifying a custom filter like `(|(mail={username})(userPrincipalName={username}))`. I've documented this in the new help page.

PR #330, opened 2017-11-30
Fix display of quoted lines containing just "0"

inguin:
When highlighting quoted lines in internal notes, lines containing
nothing except the quote marker are displayed as empty. Since the check
evaluates the line contents as boolean it mistakenly identifies the
string "0" as an empty line.


glensc:
huh. thanks.

PR #331, opened 2017-12-14
use symfony/filesystem for file writes

glensc:


glensc:
@balsdorf dropped one method from CAS backend, seemed unused.

balsdorf:
@glensc I believe that was there when I was going to have an admin page for configuring it, but it is not used now so good to remove it.

PR #347, opened 2018-02-10
Added support for "checkbox" custom fields on advanced search page

balsdorf:
Not sure why this was never implemented before. It just reuses the code path for multiple select.


glensc:
merging. assume you tested this :)

PR #357, opened 2018-03-14
Track composer.lock

glensc:
From composer docs:

- projects: [keep lock](https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control) in projects where it really matters (production is deployed from it)
- libraries: [do not keep lock](https://getcomposer.org/doc/02-libraries.md#lock-file) for libraries

This will make installation of dependencies faster, especially when switching branches with different dependencies.

There are few things required to be in place, for this to work:
- `config.platform.php` must be set to lowest version supported. this ensures resolvable deps even if developer has newer php version.
- `composer.json` may not be updated manually. use `composer require` or `composer remove` commands. if you forgot to do that, use `composer update --lock`. there may not be commit merged when `composer.json` is updated but `composer.lock` is not.

cc @balsdorf

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,7 +31,6 @@ Session.vim
 /cli/*.phar
 /cli/build/
 
-/composer.lock
 /vendor
 
 /config/cas.php
--- a/bin/ci/release.sh
+++ b/bin/ci/release.sh
@@ -129,7 +129,6 @@ clean_whitespace() {
 composer_install() {
 	# this dir does not exist in git export, but referenced in composer.json
 	install -d tests/src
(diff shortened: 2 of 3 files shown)


glensc:
@balsdorf wave or something that you got this before i merge :)

glensc:
note: this breaks php 7.2 and php 7.3 in travis because `defuse/php-encryption` 1.x requires `ext-mcrypt` which is not available.

glensc:
https://github.com/eventum/eventum/pull/358

balsdorf:
Sounds good to me. I was just discussing composer.lock today with a colleague.

Issue #350, opened 2018-03-02
Eventum auto-generated mail send out error from process_mail_queue.php


astabing:
Hi. 
I`ve got Eventum process_mail_queue.php error which looks like a bug as it triggered by Eventum own auto-generated message send to user.  We are using Eventum v3.3.4-1

Below is the error output from process_mail_queue.php:

```
sudo -u http php /usr/share/eventum/bin/process_mail_queue.php
app.ERROR: Zend\Mail\Transport\Smtp transport expects either a Sender or at least one From address in the Message; none provided {"traceFile":"/var/log/eventum/zf-mail-5a996d8404260.json","exception":"[object] (Zend\\Mail\\Transport\\Exception\\RuntimeException(code: 0): Zend\\Mail\\Transport\\Smtp transport expects either a Sender or at least one From address in the Message; none provided at /usr/share/php/Zend/Mail/Transport/Smtp.php:284)"} {"file":"/usr/share/eventum/src/Mail/MailTransport.php","line":68,"class":"Eventum\\Mail\\MailTransport","function":"send","memory_usage":"7.75 MB","memory_peak_usage":"7.75 MB","version":"3.3.4","usr_id":""}
Mail_Queue: issue #: Can't send mail 1214939: Zend\Mail\Transport\Smtp transport expects either a Sender or at least one From address in the Message; none provided
```

And here is related error trace output:

```
cat /var/log/eventum/zf-mail-5a996d8404260.json
["User Name <user.name@example.com>",{"Date":"Thu, 01 Mar 2018 17:25:14 +0200","Subject":"Eventum: User account information updated","To":"User Name <user.name@example.com>","MIME-Version":"1.0","Content-Type":"text\/plain; charset=\"UTF-8\"","Content-Transfer-Encoding":"8bit","Message-ID":"<eventum.md5.uffermije.d6ehloxka20ok@eventum.example.com>","Precedence":"bulk","Auto-Submitted":"auto-generated"},"Your user account has been updated in DELFI Eventum.\n\nYour account information as it now exists appears below.\n\n----------------------------------------------------------------------\n        Full Name: User Name\n    Email Address: user.name@example.com\n            Group: \nAssigned Projects: Project1: Reporter\n                   Project2: Standard User\n                   Mdisain: Standard User\n\n----------------------------------------------------------------------\n\n<https:\/\/eventum.example.com\/>"]
```

Related error entry from eventum.mail_queue database table:

```
[...]

PR #358, opened 2018-03-14
update to defuse/php-encryption 2.1

glensc:
[php-encryption/2.0.0](https://github.com/defuse/php-encryption/releases/tag/v2.0.0):
> Release of version 2.0.0 of defuse/php-encryption. This is a significant compatibility-breaking improvement over version 1.2.1. See the README for details.

NOTE: as the encrypted results are hex encoded by defuse/php-encryption, instead of base64, the result db columns may need to be increased. 1 byte encryption resulted 170 bytes encrypted result and 255 byte got exploded to 678 bytes.

so therefore not using defuse encoding, but keeping our own base64 implementation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ NOTE: Mail related Workflow methods changed signature, see #263
 - Use filename-normalizer for attachment filenames extracted from emails (@glensc, #356)
 - Fix Open Redirect vulnerability found by NetSparker (https://www.netsparker.com/, @balsdorf)
 - Add experimental markdown rendering (@glensc, #291)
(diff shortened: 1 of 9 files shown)


glensc:
the db migration: as there's no way to encrypt using old version, should `down` method just disable encryption?

@balsdorf

balsdorf:
Sounds like that is the only option. I'm okay with it.

glensc:
i think will delay merge of this, there are too many changes in 3.4.0 release.

have you yourself switched to code currently in master? i have tried mail sending, and have disabled several times due some issues. #353, #352, #350, #351 all somewhat related, those are solved now. but haven't tested in production download_emails.php part much either.

balsdorf:
I have not yet but I will next week.

PR #364, opened 2018-04-25
Fix bug adding sender to authorized repliers list when unblocking email

balsdorf:
So in [Note::convertNote](https://github.com/eventum/eventum/blob/4e95a61ee6df5c6c8a1fabc5e749e00ea582ab66/lib/eventum/class.note.php#L579) the `$mail` object is passed to `Notification::notifyNewEmail` which changes the sender. Then on [line 588](https://github.com/eventum/eventum/blob/4e95a61ee6df5c6c8a1fabc5e749e00ea582ab66/lib/eventum/class.note.php#L588) `$mail->from` is accessed which now shows the sender from Notification::notifyNewEmail. 

I've now made `notifyNewEmail` create a copy of the mail object before modifying it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Eventum 3.4.x requires PHP 5.6.
 - Fix bug with project switcher not redirecting to previous page (@balsdorf)
 - Fix bug dumping routed mail to save directory (@balsdorf)
 - Update cebe/markdown to 1.2.1 (@glensc, #365)
+- Fix bug with adding sender to authorized repliers list when unblocking an email (@balsdorf, #364)
 
 [3.4.1]: https://github.com/eventum/eventum/compare/v3.4.0...master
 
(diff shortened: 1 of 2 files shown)


glensc:
don't need heavy stringfy and parse again. `clone` should work:

https://github.com/eventum/eventum/blob/v3.4.0/lib/eventum/class.notification.php#L425

balsdorf:
Rebased.

I thought clone would work as well, but it does not. The $headers object contains a list of objects by reference that are not cloned. 

I can modify `Zend\Mail\Storage\Part` to implement the `__clone()` method which solves this problem and submit a PR to https://github.com/glensc/zend-mail/tree/develop-2.4 if you like.

glensc:
ok. lets have it this way for now. but yes. submit it, but preferrably to upstream: https://github.com/zendframework/zend-mail/ so i could just backport it.

altho half of the clone functionality is in eventum mailmessage class, that needs to be implemented too.

and in any case, if upstream rejects the change, there's no point maintaining it in own version of zend-mail .

glensc:
someone reported this upstream too https://github.com/zendframework/zend-mail/issues/214

PR #367, opened 2018-04-30
Fix parsing of link references in markdown

glensc:
Fix parsing of link references in markdown.

https://github.com/cebe/markdown/issues/157

### DIFF
--- a/lib/eventum/class.link_filter.php
+++ b/lib/eventum/class.link_filter.php
@@ -263,7 +263,9 @@ class Link_Filter
             $parser->enableNewlines = true;
         }
 
-        $text = $parser->parseParagraph($text);
+        $text = $parser->parse($text);
+        // strip paragraph, confuses single line areas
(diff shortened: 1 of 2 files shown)


lib/eventum/class.link_filter.php
```diff
@@ -263,7 +263,9 @@ public static function markdownFormat($text)
             $parser->enableNewlines = true;
         }
 
-        $text = $parser->parseParagraph($text);
+        $text = $parser->parse($text);
+        // strip paragraph, confuses single line areas
+        $text = preg_replace("{^<p>(.+)</p>\n$}", '$1', $text);
```
cebe: note, that this will allow creating markdown that has `</p><p>` inside which results in invalid HTML. You should probably have two different `markdownFormat()` methods, one for "single line" context and another for longer texts.

glensc:
going to merge this as is for now

PR #368, opened 2018-04-30
Decouple IRC logic to events

glensc:
Intention is to replace IRC with different implementation, post to Slack for example.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Upgrading to 3.4.x versions requires that you upgrade to 3.2.0 version first.
 Eventum 3.4.x requires PHP 5.6.
 
 - XMLRPC: Add methods to support eventum/irc-bot over xmlrpc (@glensc, #369)
+- Decouple IRC logic to events (@glensc, #368)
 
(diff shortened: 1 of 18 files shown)


balsdorf:
YES! I have been wanting to do this. Currently we have an ugly hack to send events to slack instead of IRC but proper support would be great.

glensc:
i've added events that my workflow handled for now.

i've split logic to two classes, `IrcSubscriber` and `IrcLegacySubscriber`. `IrcLegacySubscriber` can be dropped in the future (edit `config/setup.php` `extensions` list, if you want to do it now)

IrcSubscriber will be kept, to support adding data to irc_notice table.

if nothing calls `Notification::notifyIRC()` then the `irc_notice` table is not filled, currently only workflow classes may call `Notification::notifyIRC()`, so it will be noop if integration not used.

i've added my irc integration to docs/examples/extension/Event/IrcSubscriber.php as an example.

do you see how you could add your slack integration top of this branch, are the pieces needed in place?

glensc:
created new project for irc bot too, so it's code could be dropped from main repo: https://github.com/eventum/irc-bot/pull/1

PR #369, opened 2018-05-01
XMLRPC: Add methods to support eventum/irc-bot over xmlrpc

glensc:
xmlrpc method changes:

- isValidLogin checks for user status as well
- new: getClockedInList - unrestricted
- new: getQuarantinedIssueList - unrestricted
- new: getProjectByName - unrestricted
- new: getPendingMessages - requires ROLE_ADMIN
- new: markEventSent - requires ROLE_ADMIN

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 Upgrading to 3.4.x versions requires that you upgrade to 3.2.0 version first.
 Eventum 3.4.x requires PHP 5.6.
 
+- XMLRPC: Add methods to support eventum/irc-bot over xmlrpc (@glensc, #369)
 
 [3.4.2]: https://github.com/eventum/eventum/compare/v3.4.2...master
 
--- a/db/migrations/20180501205449_eventum_irc_notice_indices.php
+++ b/db/migrations/20180501205449_eventum_irc_notice_indices.php
@@ -1,0 +1,25 @@
+<?php
+
+/*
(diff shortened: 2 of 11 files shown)


glensc:
@balsdorf please have your input about new methods, needed for the standalone irc bot: https://github.com/eventum/irc-bot/pull/1

glensc:
i updated PR body, there are more changes now.

balsdorf:
I don't think that will be a problem, however I'm also fine with changing the method to only show the specific data we need right now and add additional fields as needed in the future.

glensc:
there's problem with clockIn method. it performs as current user. but irc-bot is designed to use admin user. 

i had an idea to add `sudoTimeClock($usr_id, $action)` method, but currently to verify calling user role, i need project id (see the diffs in this PR), but the clock irc commands have no project context. should i just hardcode project id to `1`?

balsdorf:
Internally we have APP_DEFAULT_PROJECT which I use when I have to hard code a project ID, though it is just 1. I think either a constant or just hardcoding 1 is an acceptable solution for now.

glensc:
Some other idea i had, is to cache user login+token when they do `!auth` irc command, and use these to do xmlrpc calls.

but the system user for xmlrpc is still needed, as the events polling has no user context.

glensc:
Went with switch user at client side, as the credentials are asked already from user: https://github.com/eventum/irc-bot/pull/3

PR #397, opened 2018-07-06
update composer deps

glensc:
- Updating paragonie/random_compat (v2.0.15 => v2.0.17)

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1229,16 +1229,16 @@
         },
         {
             "name": "paragonie/random_compat",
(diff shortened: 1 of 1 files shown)


glensc:
btw, perhaps we should drop `Roave/SecurityAdvisories`. has no value and creates just diff when updating lockfile

@balsdorf

balsdorf:
@glensc Agree on dropping `Roave/SecurityAdvisories`. Was looking at the lockfile and the diff wasn't clean.

PR #398, opened 2018-07-12
Handle uncaught header exception from mail download

glensc:
the uncaught exception made one faulty email stop email download process

```
In Address.php line 41:

The input is not a valid email address. Use the basic format local-part@hostname
```

the faulty email was:
```
From: "Benny   Higgins " <>
```


glensc:
currently just skips the mail. thus it will continue spam cron until such email is removed manually from mailbox.

at least it doesn't break processing other mails.

ideas what else could be done are welcome

cc @balsdorf

PR #406, opened 2018-08-28
Mail: Rename unparseable headers

glensc:
This makes it possible to still load emails, whose headers can't be parsed.

The headers are named with `X-Broken-Header-` prefix


balsdorf:
WOOO! I've been meaning to do something like this for a long time, but renaming them to X-Broken is something I hadn't thought of.

glensc:
This will only work for headers that are not `GenericHeader`, i.e in this case `Cc:` header.

PR #410, opened 2018-08-29
Strip non numerical input from issue lookup shortcut

balsdorf:
Currently if you had any non numeric characters such as whitespace or "." in the issue lookup shortcut field results in an error message. This automatically strips any non numeric values from the input.

### DIFF
--- a/htdocs/js/main.js
+++ b/htdocs/js/main.js
@@ -61,8 +61,9 @@ $(document).ready(function() {
 
     $('#shortcut_form').submit(function(e) {
         var target = $('#shortcut');
-        if (!Validation.isNumberOnly(target.val())) {
-            alert('Please enter numbers only');
+        target.val(target.val().replace(/\D/g,''));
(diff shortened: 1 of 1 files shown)


glensc:
I thought this was always there.

LGTM; add changelog entry too

htdocs/js/main.js
```diff
@@ -61,8 +61,9 @@ $(document).ready(function() {
 
     $('#shortcut_form').submit(function(e) {
         var target = $('#shortcut');
-        if (!Validation.isNumberOnly(target.val())) {
-            alert('Please enter numbers only');
+        target.val(target.val().replace(/\D/g,''));
+        if (Validation.isWhitespace(target.val())) {
```
glensc: but `target` will be empty, no need for `Validation.isWhitespace()` call

also, for DRY, create just `var value = target.val().replace(/\D/g,'')` instead of calling `target.val()` repeatedly?

PR #422, opened 2018-11-29
Add SystemEvents::hasEvent to feature test availability

glensc:
This is useful from workflow to test different eventum versions.

It's quite common for me to pre-test some eventum branch while using same workflow setup.

cc @balsdorf


glensc:
while at it, moved some internal classes to sub namespace. not related to the change

PR #423, opened 2018-12-19
WIP: Handle broken Sender header

glensc:


glensc:
cc @balsdorf

glensc:
@balsdorf the problem is that somehow users managed to add eventum own address (`issue-xxx@example.net`) to issue notification list, and then every note/email sent from issue created email which has empty `Sender: ` header.

i could not even find place in code what sets `Sender: ` header... any ideas?

balsdorf:
I don't think there is anything in Eventum that explicitly sets `Sender: ` so it must be set either in the mail library or by the server.

When I send a test email, I don't see `Sender:` in the mail queue.

PR #428, opened 2018-12-26
WIP: Escape HTML in plain description.

balsdorf:
Previously, putting an open comment would break the rest of the page due to plain description not being escaped.

```
<!-- everything after this is broken
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 Upgrading to 3.5.x versions requires that you upgrade to 3.2.0 version first.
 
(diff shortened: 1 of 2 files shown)


balsdorf:
Forgot to put WIP on this. Had to apply an emergency fix internally for this, haven't had time to fully look at it. Forgot about markdown mode as we don't use that internally, but a brief test shows it explodes horribly with this test string when markdown mode is enabled.

```
<hello Version="V1.0.0">
<!-- Danging comment
<blah>64GB</blah>
</hello>

**Bold**
```

glensc:
looks like the same fix was made in #515

Issue #412, opened 2018-09-06
problems erasing attachments


lmoralesa:
Hi, we did and upgrade from eventum 3.2.0 to 3.5.3, in version 3.2.0 users did have permissions to erase attachments, now I am receiving reports that users can not erase attachment.
I need help to see if you can guide me in this issue

glensc:
You really need to give more information, can't just guess what might be wrong.

lmoralesa:
for example, a user creates an issue, fills the fields with information on the issue, attaches a * .txt or * .pdf file. and the issue is created. Once the issue is created, it is detected that the attached file is not correct, it is about removing the attachment and eventum gives an error message that says "You do not have the permission to remove this attachment."
The funny thing is that the message is given to the same user who attached the file, if it guides me where to obtain a log to send it and help us analyze

lmoralesa:
I have and user that attached some files, he tries to erase them but got an error.

![image](https://user-images.githubusercontent.com/23003333/47865428-ce0a2c80-ddc1-11e8-8511-8d2d439dd414.png)

glensc:
what is the role of user who is trying to delete? how where attachments created? via issue create via email or via web interface?

lmoralesa:
The user has standart role, but we change the user to admin role and still can not erase the attach, the  attach files are upload via web, files are .pdf and .docx

glensc:
was able to reproduce:
- upload with user role "user" and set min role as "standard user"
- attempt to delete such attachments results "-2"
- uploading with min role lower than that, works okay

this corresponds to these code paths:
- https://github.com/eventum/eventum/blob/v3.5.5/src/Attachment/AttachmentManager.php#L409
- https://github.com/eventum/eventum/blob/v3.5.5/src/Attachment/AttachmentManager.php#L381


i found also funny bug:
- user with role "user" was able to pick userlevel bigger from their own. the ui displayed success and closed the popup.

cc @balsdorf

PR #430, opened 2018-12-26
Fix attachment delete permissions check

glensc:
fixes: #412

the problem was caused by mispatched less than operator
```diff
- ($usr_id != $group->user_id && User::getRoleByUser($usr_id, Issue::getProjectID($group->issue_id) < User::ROLE_MANAGER))
+ ($usr_id != $group->user_id && User::getRoleByUser($usr_id, Issue::getProjectID($group->issue_id)) < User::ROLE_MANAGER)

```

the bugs were introduced by:
- ff310a91c8a508e5e3c68b287fce364f91e0e4d1
- 1d49286a35af4c049daef16584a295ddb8a1c1b6

simplified code to use local variables

cc @balsdorf, @lmoralesa

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Upgrading to 3.5.x versions requires that you upgrade to 3.2.0 version first.
 
 - Extensions: skip components that fail to initialize (@glensc, #431)
 - Handle text/html only email rendering (@TexasGitHubber, @glensc, #399, #429, #400, #401)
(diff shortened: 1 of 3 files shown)


glensc:
@balsdorf added link to commits that introduced the bug. please review if the original intent is still there.

balsdorf:
Thanks @glensc . The old behavior was not intended, surprised it was missed this long.

PR #438, opened 2019-01-09
add SystemEvents::NOTIFICATION_NOTIFY_ADDRESS event handling

glensc:


glensc:
@balsdorf what you think?

glensc:
probably this should be handled differently in the future (when workflow method is dropped):

- create `AddressList` to notify
- call event that would modify `AddressList`
- if `AddressList` is empty, there's nothing to iterate over

PR #439, opened 2019-01-09
Detect invalid mail header starting with "From " and replace it with a correct custom one.

vladsf:
Email viewer (view_email.php) fails with an exception "Invalid header name detected"
if a stored raw email starts with a Mbox line "From name@example.com .." which is
not a correct mail header.

Sample email tests/data/from_nocolon.txt may be used as a test case.

### DIFF
--- a/src/Mail/Helper/MailLoader.php
+++ b/src/Mail/Helper/MailLoader.php
@@ -118,6 +118,11 @@ class MailLoader
                     $line = str_replace("\r", '', $line);
                 }
 
+                if ($e->getMessage() === 'Invalid header name detected' && substr($line, 0, 5) === 'From ') {
+                    // remove Mbox header from message if it is present
+                    $line = "X-Broken-Header-Mbox: $line";
+                }
+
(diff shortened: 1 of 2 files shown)


vladsf:
> `tests/data/from_nocolon.txt` is missing

the file is in the master branch already, I did not want to add a similar copy sample:
https://github.com/eventum/eventum/blob/master/tests/data/from_nocolon.txt

vladsf:
> please fill actual title for the PR. "fix issue" is too ambiguous...

done.

glensc:
ah, i did not realize that file already existsed. can you please add testcase using that file via your modified code path?

vladsf:
> 
> 
> ah, i did not realize that file already existsed. can you please add testcase using that file via your modified code path?

OK, I will add a testcase function to tests/Mail/LoadEmailTest.php.

glensc:
no. why would you create new one? you add more commits to existing PR. thus this one: #439

vladsf:
> no. why would you create new one? you add more commits to existing PR. thus this one: #439

My mistake. Fixed. Added new commit to this PR.

Issue #427, opened 2018-12-25
"Invalid header name detected" (From XXX) on upgrades


vladsf:
Hi,

I have "Invalid header name detected" issue on 'php bin/upgrade' running and in browser viewing older issues e.g. http://eventumtest01.test.com/view_email.php?ema_id=20&id=236398

I saw #155 issue, but the fix does not solve the trouble on upgrades or views. 

Eventum v3.1.10 - php bin/upgrade:
```
[2018-12-24 15:44:05] db.INFO: skipped maq_id=867359, exception: Line "From xxx.xxx@company.com  Mon Jan  5 13: 05:51 2015"does not match header f
ormat! [] {"file":"/intranet/eventum-3.1.10/upgrade/patches/56_maq_message_id.php","line":52,"class":"Eventum\\Db\\Migrate","function":"{closure}","memory_
usage":"68.5 MB","memory_peak_usage":"74 MB","version":"3.1.10","usr_id":""}
```
The same on upgrade run from 3.2.0 to 3.5.4:
```
[2018-12-25 10:14:15] db.INFO: skipped maq_id=867198, exception: Invalid header name detected [] {"file":"/intranet/eventum-3.5.4/db/migrations/20180306202659_eventum_maq_message_id.php","line":49,"class":"EventumMaqMessageId","function":"up","memory_usage":"42.75 MB","memory_peak_usage":"45.25 MB","version":"3.5.4","usr_id":""}
```
Then it fails in browser (shows empty page) and eventum.log gets: 
```
[...]

PR #440, opened 2019-01-11
Detect invalid mail header starting with "From " (issue #427) and a testcase

vladsf:
Email viewer (view_email.php) fails with an exception "Invalid header name detected"
if a stored raw email starts with a Mbox line "From name@example.com .." which is
not a correct mail header.

This PR includes a testcase.


glensc:
what was wrong with previous PR? please close duplicates yourself!

PR #441, opened 2019-01-15
Allow auth adapters to be chained

glensc:
Allow auth adapters setup from config:

the default:
```php
    'auth' => [
        'adapter' => MysqlAdapter::class,
    ],
```

with chain adapter with ldap and mysql:
```php
    'auth' => [
        'adapter' => ChainAdapter::class,
        'options' => [
            ChainAdapter::class => [
                0 => [
                    LdapAdapter::class,
                    MysqlAdapter::class,
                ],
            ],
        ],
    ],
```

## tasks

- [x] drop `APP_AUTH_BACKEND` and `APP_AUTH_BACKEND_ALLOW_FALLBACK`; add migration
- [ ] (optional) strict typing to new classes
- [x] move out `LdapAdapter::getDefaults`
- [x] test `LdapSyncCommand`
- [x] manage (admin interface): ensure it's usable (and help page too)
- [ ] check what redundant, duplicate methods could be moved out of adapter (traits, other classes?)

### DIFF
--- a/lib/eventum/class.auth.php
+++ b/lib/eventum/class.auth.php
@@ -11,6 +11,8 @@
  * that were distributed with this source code.
  */
 
+use Eventum\Auth\Adapter\AdapterInterface;
+use Eventum\Auth\Adapter\MysqlAdapter;
 use Eventum\Monolog\Logger;
 use Eventum\Session;
 use Symfony\Component\Filesystem\Exception\IOException;
@@ -482,21 +484,21 @@ class Auth
     }
 
     /**
-     * @return Auth_Backend_Interface
+     * @return AdapterInterface
      */
     public static function getAuthBackend()
     {
-        /** @var Auth_Backend_Interface $instance */
+        /** @var AdapterInterface $instance */
         static $instance = false;
 
-        if ($instance == false) {
+        if ($instance === false) {
             $class = APP_AUTH_BACKEND;
 
             // legacy: allow lowercase variants
-            if (strtolower($class) == 'mysql_auth_backend') {
-                $class = 'Mysql_Auth_Backend';
-            } elseif (strtolower($class) == 'ldap_auth_backend') {
-                $class = 'LDAP_Auth_Backend';
+            if (strtolower($class) == 'Eventum\Auth\Adapter\MysqlAdapter') {
+                $class = 'Eventum\Auth\Adapter\MysqlAdapter';
+            } elseif (strtolower($class) == 'Eventum\Auth\Adapter\LdapAdapter') {
+                $class = 'Eventum\Auth\Adapter\LdapAdapter';
             }
 
             try {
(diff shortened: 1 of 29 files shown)


glensc:
@balsdorf this is initial design, it's supposed to be configurable which adapters I want to chain. or use adapter directly.

please go over the methods and give your thoughts.

i need to enable ldap+mysql for some grace period, and later drop mysql part, and current fallback method is not working, so decided to rewrite the auth adapter part.

balsdorf:
I will check it out this week.

lib/eventum/class.auth.php
```diff
@@ -481,60 +482,29 @@ public static function setCookie($name, $value, $expiration)
...
+        if ($adapter === false) {
+            $spec = Setup::get()['auth'] ?? [];
 
             try {
-                $instance = new $class();
-            } catch (AuthException $e) {
-                $message = "Unable to use auth backend '$class'";
+                $adapter = Eventum\Auth\Adapter\Factory::create($spec);
+            } catch (Throwable $e) {
+                $message = 'Unable to use instantiate auth adapter';
```
balsdorf:
```suggestion
                $message = 'Unable to instantiate auth adapter';
```

balsdorf:
This looks great. Much cleaner then the old process. Just had the one error message to change :+1:

glensc:
some chained methods do not make sense, may end up removing them, or throw an exception

glensc:
some other cleanup needed too, for example ldap adapter calls Auth::saveLoginAttempt, while mysql adapter doesn't. and what if they are chained, these should be logged at most top level call; if any...

glensc:
the admin interface is more or less usable, allows to choose adapter, but not really configure it.

making it perfect is too much of an effort with current templates.

glensc:
@balsdorf this satisfies my goals (can login with mysql and ldap); give your LGTM for merge.

the rest of the ideas can be implemented later (if ever ;)...)

balsdorf:
I have not had time to test this, but looking at the design it looks good. I agree that we should split some things out later, but I'm happy for you to merge this.

glensc:
@balsdorf please test, does this break that legacy cli app, as it bundles `class.auth.php`, but made changes to that class in this PR scope.

not fun maintaining that legacy piece: https://github.com/eventum/eventum/pull/451

balsdorf:
Legacy CLI works for me with default MySQL Authentication. Have not had time to test other auth methods yet.

glensc:
Auth setting is server side, should not affect the CLI.

Btw, what version you are using for CLI? as I tested with 7.1 and got a bunch of notices on the terminal.

balsdorf:
CLI worked fine with 7.1 and 7.3. Server side I had to suppress errors as
there was a warning.

Honestly, I'm fine with removing the classic CLI. There are a few people
who use it internally who will be cranky but I don't think there is
widespread public use. If there is our time is probably better spent on the
new CLI then updating the old one.

glensc:
https://github.com/eventum/eventum/pull/453

but they can keep using the cli from last release, there were hardly any changes to cli app anyway.

Issue #138, opened 2016-02-04
change installation to ask for default user and password


glensc:
currently there are two users created to system (one inactive)
both hashes can be resolved quickly via rainbow tables from internet search

best is to solve this by creating admin user in setup screen with password input by user. that should randomize quite a bit the used password and use latest provided password hashing, not md5

ps: https://bugs.mysql.com/bug.php?id=7551

glensc:
the password is hashed using password hash, making the result not findable with rainbow tables: https://github.com/eventum/eventum/pull/444

however the password is still well known...

cc @balsdorf

PR #444, opened 2019-01-16
Drop MD5 and MD5-64 password support

glensc:
Passwords are hashed using `password_hash` since `3.0.4` (released on 2015-10-31)

Support for MD5 and MD5-64 hashes were dropped in 

- New passwords are saved using more secure hashing than before (#77)
- Upgrade password hash on successful login (#93)

Such old passwords are either stale or rehashed on login.

## Tasks

- [x] check that installation uses compatible hash

cc @balsdorf


glensc:
TODO: check that on install admin password is saved using proper hash, may need UI change to ask for initial password.

glensc:
using `AuthPassword::hash` for now.

issue to ask with ui the initial password already exists:
- https://github.com/eventum/eventum/issues/138

PR #445, opened 2019-01-16
Secure unserialize: control allowed_classes options

glensc:
PHP 7.0 adds `allowed_classes` element of options to be able to specify which classes may be instantiated.

See:
- http://php.net/unserialize

# test
- [x] test `DateTime` objects with XMLRPC

cc @balsdorf


glensc:
tested with cli weekly-reports command

Issue #447, opened 2019-01-23
Mentioned on Feature


glensc:
I'd like to have `mentioned on` feature in Eventum

Similar to what JIRA has:
- https://developer.atlassian.com/server/jira/platform/jira-rest-api-for-remote-issue-links/
- https://developer.atlassian.com/server/jira/platform/using-fields-in-remote-issue-links/

glensc:
I started https://github.com/eventum/eventum/pull/448 to create the mentioned events from gitlab events.

@balsdorf: ideas of structure and on the naming?

i was thinking to start with jira like structure:
- table: `remote_links(int rel_iss_id, string rel_gid, string rel_relationship, string rel_url, string rel_title)`?

glensc:
In #448 I just emit EventDispatcher event, so i can have workflow method calling API to insert records. The API in this case would be just Doctrine Repository methods.

PR #449, opened 2019-01-23
Add backend and frontend render for mentioned on feature

glensc:
This adds `remote_link` table, API and UI, see #447

Filling the data is to be done by extensions: #448

The `remote_link` table:

```mysql
CREATE TABLE `remote_link` (
  `rel_id` int(10) unsigned NOT NULL AUTO_INCREMENT,
  `rel_iss_id` int(11) unsigned NOT NULL,
  `rel_gid` varchar(255) CHARACTER SET ascii DEFAULT NULL COMMENT 'Global Id',
  `rel_relation` varchar(255) CHARACTER SET ascii NOT NULL COMMENT 'Link relationship type',
  `rel_url` text CHARACTER SET ascii NOT NULL,
  `rel_title` varchar(255) NOT NULL,
  PRIMARY KEY (`rel_id`),
  KEY `rel_id` (`rel_id`,`rel_gid`)
) ENGINE=MyISAM DEFAULT CHARSET=utf8
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ first, also the PHP version required is 7.1, see [#426].
 - Remove Legacy CLI application (@glensc, #453)
 - Move Login-Back-Off logic to MySQL adapter only (@glensc, #455)
 - Use `symfony/var-exporter` for writing config files (@glensc, #456)
+- Add support to fill Mentioned on data (@glensc, #447, #449)
 
 [3.6.0]: https://github.com/eventum/eventum/compare/v3.5.6...master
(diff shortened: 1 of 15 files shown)


glensc:
decided to use `ascii` for for url and relationship, as these should not be anything else

glensc:
@balsdorf the api and ui part is ready.

1. i wonder what should be visibility of these. currently block is show if links present
2. how to order the links. currently database order is used
3. there's no way to remove the links from the ui. if remove support is added, what user level? `ROLE_USER`?

glensc:
i wonder, should the records have some timestamp in them...

glensc:
merging then

PR #451, opened 2019-01-23
cli: use pld distfiles for XML_RPC pear (and some more) packages

glensc:
pear.php.net is down:
- https://travis-ci.org/eventum/eventum/jobs/483566768<Paste>

apparently comproposed, see @pear twitter post:
- https://twitter.com/pear/status/1086634389465956352

and so are unavailable all it's mirrors:
https://web.archive.org/web/20180904190513/https://pear.php.net/mirrors.php

- http://pear.php.net/ - points to @cweiske site and that site no content
- http://us.pear.php.net/ - dns missing
- http://de.pear.php.net/ - ip not responding, and maintainer also being @cweiske

and there is no packagist version of that pear package:
- https://packagist.org/search/?q=pear/xml_rpc

so, embed package resource to composer.json from pld distfiles

this cli is legacy we carry around as the new cli is not that
functional:
- https://github.com/eventum/cli/issues/2


glensc:
https://news.ycombinator.com/item?id=18987518

PR #455, opened 2019-01-27
Move Login-Back-Off logic to MySQL adapter only

glensc:
Move Login-Back-Off logic to MySQL adapter only: It makes no sense for other adapters than MySQL and they've implemented methods as dummy anyway. The external systems likely account themselves such logic.

The major concern was that the accounting was `usr_id` based, meaning unknown users are not accounted, to resolve that would need another table, which seems too much for now.

This also moves constants to setup config:
- `APP_FAILED_LOGIN_BACKOFF_COUNT`
- `APP_FAILED_LOGIN_BACKOFF_MINUTES`

refs:
- https://github.com/eventum/eventum/commit/780ddaa822866c432b97d92adfff3934a70eaa01
- https://github.com/eventum/eventum/pull/441#issuecomment-455273425
- https://github.com/eventum/eventum/pull/441#issuecomment-456530674

cc @balsdorf

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ first, also the PHP version required is 7.1, see [#426].
 - TextMessage: do not parse subparts if no content-type given (@glensc, #452)
 - Allow auth adapters to be chained (@glensc, #441)
 - Remove Legacy CLI application (@glensc, #453)
+- Move Login-Back-Off logic to MySQL adapter only (@glensc, #455)
 
 [3.6.0]: https://github.com/eventum/eventum/compare/v3.5.6...master
 [#426]: https://github.com/eventum/eventum/pull/426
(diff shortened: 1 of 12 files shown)


glensc:
@balsdorf as this changes bits authentication, please review! I already broke one case where `bool` cast converted `3` to `1`.

glensc:
I would not want to show any error at all than "authentication failed", even remembering username is information leakage (it remembers valid user)

but that can be done in a separate PR.

glensc:
uh, the failed login count keeps incrementing for the app system user here

```
select usr_failed_logins from user where usr_id=16;
+-------------------+
| usr_failed_logins |
+-------------------+
|            307263 |
+-------------------+
1 row in set (0.00 sec)
```

it's probably mail download cron.

but i don't even have the feature enabled:

```php
            'login_backoff' => [
                'count' => null,
                'minutes' => 15,
            ],
```

Issue #262, opened 2017-05-11
Performance: Option to disable Issue Navigation.


lauripiisang:
### Description
The current Issue Navigation (Previous Issue , Next Issue) uses your search filter.
If your last search was slow, your every page load will be slow.
It isn't obvious to the user why that is.
I don't believe all users actually require this functionality, especially when it comes at such a cost.

Account Preferences should have an option to turn off Previous/Next Issue Navigation.
Search will still be persisted, so if you go to search page you will have the same results.
Database will not be queried for next/prev issue, so user will not have slow loading problems.

Optional goodie:
Mention the option to disable the navigation with a flash message, if the user ran a slow query while loading an issue. 

### User Stories
When I navigate to an issue
And I have Issue Navigation enabled
And my next/previous issue search takes >3 seconds
And I haven't seen the message yet (cookie not set)
Then I should see a flash message
```
"You are using a slow search query with issue navigation enabled. 
If you want pages to load faster, you can disable Issue Navigation, 
but you'll lose <<< Previous Issue and Next Issue >>> navigation options. 
You can enable/disable Issue Navigation in [Link] Account Preferences
[Link] Disable Issue Navigation now
[x] Ignore"
```
When I click Ignore
Then a cookie should be set (expires in 90 days) , remembering I have already seen the notice.

When I click Disable Issue Navigation now
Then an asynchronous request should be sent to the server for disabling that preference 
And the flash message should be dismissed

When I click Account Preferences
Then I should be taken to account preferences.

balsdorf:
Personally I don't use this at all. It is leftover from the old days when people didn't use multiple tabs. Now I feel that if someone wants to navigate among issues they will open them in separate tabs.

I'm in favor of removing it completely.

@glensc Thoughts?

lauripiisang:
I'd persist it in case people have workflows set up where they traverse through tickets and don't feel like opening 50 tabs manually 

(Think of a support desk where I have a filter set up for tickets requiring my attention, I'd rather click on the first one, do all my actions, then click next. repeat.)

PR #457, opened 2019-01-29
Search profile refactor

glensc:
refs: #262


glensc:
no functional changes. merging as finishing would take longer than expected

PR #459, opened 2019-01-31
use DateTime::ATOM as DateTime::ISO8601 is not iso8601 compatible

glensc:
- https://www.reddit.com/r/ISO8601/comments/8lkcns/the_php_iso8601_date_format_is_not_actually/
- http://php.net/manual/en/class.datetimeinterface.php#datetime.constants.iso8601

the differences are:
```php
    const ISO8601 = 'Y-m-d\TH:i:sO';
    const ATOM    = 'Y-m-d\TH:i:sP';
```

http://ee1.php.net/manual/en/function.date.php#refsect1-function.date-parameters

format character | Description | Example returned values
-- | -- | --
O | Difference to Greenwich time (GMT) in hours  | Example: +0200
P | Difference to Greenwich time (GMT) with colon between hours and minutes (added in PHP 5.1.3)  | Example: +02:00

### DIFF
--- a/lib/eventum/class.date_helper.php
+++ b/lib/eventum/class.date_helper.php
@@ -160,7 +160,7 @@ class Date_Helper
         if ($omit_offset) {
             $fmt = 'Y-m-d\TH:i:s';
         } else {
-            $fmt = DateTime::ISO8601;
+            $fmt = DateTime::ATOM;
         }
 
         return $date->format($fmt);


glensc:
@balsdorf I wonder if anything will break due this?

glensc:
the only place i found the method being used was:
- https://github.com/eventum/eventum/blob/v3.5.6/lib/eventum/class.template_helper.php#L193

glensc:
which seems to be used only in post note template

```
$ grep -r 'core.current_user_datetime' templates/
templates/post_note.tpl.html:                {html_select_date start_year='-1' day_value_format='%02d' field_array='date' prefix='' time=$core.current_user_datetime all_extra=' onChange="Eventum.calcDateDiff(this.form, 1, this.name)"'}&nbsp;
templates/post_note.tpl.html:                {html_select_time minute_interval=5 field_array='date' prefix='' all_extra=' onChange="Eventum.calcDateDiff(this.form, 1, this.name)"' display_seconds=false time=$core.current_user_datetime}
templates/post_note.tpl.html:                {html_select_date start_year='-1' day_value_format='%02d' field_array='date2' prefix='' time=$core.current_user_datetime all_extra=' onChange="Eventum.calcDateDiff(this.form, 2, this.name)"'}&nbsp;
templates/post_note.tpl.html:                {html_select_time minute_interval=5 field_array='date2' prefix='' all_extra=' class="default" onChange="Eventum.calcDateDiff(this.form, 2, this.name)"' display_seconds=false time=$core.current_user_datetime}
```

https://github.com/eventum/eventum/blob/v3.5.6/templates/post_note.tpl.html

glensc:
and with `true` parameter, so the affected code path is never used! ...

PR #464, opened 2019-02-04
handle authorized_replier removal error with no user ids

glensc:


glensc:
looks there are more assumptions that `removeRepliers` is called on single id:
- https://github.com/eventum/eventum/blob/v3.5.6/lib/eventum/class.authorized_replier.php#L97

glensc:
changed method to take issue id as parameter, and remove that previous nonsense.

cc @balsdorf

PR #466, opened 2019-02-05
Refactor Custom Fields to use Interfaces and Doctrine Models

glensc:
This attempts to bring some clarity to custom fields structure. It's always been a mystery to me.

Currently, this PR doesn't do anything useful, other trying to preserve compatibility.

Backward incompatibility: If you extend `Dynamic_Custom_Field_Backend`, you need to update to match methods signature.

cc @balsdorf


glensc:
Some bug appeared (ok in master)

So, If I add new custom field, and add it to listing, then listing is broken because there's missing `<td>` for older issues, because they have no value from sql join.

image

that's from `Search::getListing()`, funny is, that i didn't modify that part...

glensc:
oh, the old code used left join...

https://github.com/eventum/eventum/blob/v3.5.6/lib/eventum/class.custom_field.php#L717-L726

glensc:
https://github.com/eventum/eventum/blob/v3.6.0/lib/eventum/class.custom_field.php#L84-L92

this looks broken, it tries to `delete from issue_custom_field where icf_value=$cfo_ids`, but icf_value does not contain issue_custom_field id, but actual values...

@balsdorf what do you think?

glensc:
okay, the logic was added in initial commit f51c76902f, and seems really, the option values are stored in `icf_value` fields in `issue_custom_field` table, even how odd it is not. those might be converted to text values by custom fields. so it's really confusing. i hope to get more stable api, when i'm through converting the old methods and then i'll understand the logic more.

glensc:
i probably would add new column to `issue_custom_field`, when value is id of custom_field_option, store it into new `icf_option_id` column, then the type would always be correct, as currently there's chance to have string value when trying to lookup `custom_field_option` based on `icf_value`

```

mysql> select * from issue_custom_field order by icf_fld_id;
+--------+------------+------------+----------------------+-------------------+----------------+
| icf_id | icf_iss_id | icf_fld_id | icf_value            | icf_value_integer | icf_value_date |
+--------+------------+------------+----------------------+-------------------+----------------+
|      1 |          1 |          1 | bo_helpdesk          |              NULL | NULL           |
|      2 |         32 |          1 | bo_helpdesk          |              NULL | NULL           |
|      5 |         20 |          1 | ads_automated_tester |              NULL | NULL           |
|      6 |         20 |          1 | adsadmin             |              NULL | NULL           |
|      7 |         20 |          1 | advertisement        |              NULL | NULL           |
|      3 |         20 |          2 | 1                    |              NULL | NULL           |
|      4 |         20 |          2 | 2                    |              NULL | NULL           |
|     15 |         22 |          3 | 16                   |              NULL | NULL           |
|     10 |         54 |          3 |                      |              NULL | NULL           |
|     14 |         22 |          3 | 15                   |              NULL | NULL           |
|     16 |         22 |          3 | textare              |              NULL | NULL           |
+--------+------------+------------+----------------------+-------------------+----------------+
11 rows in set (0,00 sec)
```

glensc:
https://github.com/eventum/eventum/blob/v3.6.0/lib/eventum/class.custom_field.php#L1704-L1745

in `updateValuesForNewType` method converting from/to date makes also no sense to me, if previous value was "16", converting it to date is "1970-01-01 00:00:16.000000"...

imho only values to recover is text to integer and vice versa....

glensc:
too many changes... will attempt to merge this soon. no real "features" yet

glensc:
the `0000-00-00` gets formatted `-0001-11-30` as in new code.

```
11:07:16 mysql{3}> select * from issue_custom_field where icf_iss_id=107622;
+--------+------------+------------+-----------+-------------------+----------------+
| icf_id | icf_iss_id | icf_fld_id | icf_value | icf_value_integer | icf_value_date |
+--------+------------+------------+-----------+-------------------+----------------+
| 154603 |     107622 |          1 | NULL      |              NULL | 0000-00-00     |
| 154599 |     107622 |         13 | 73        |              NULL | NULL           |
| 154600 |     107622 |         13 | 75        |              NULL | NULL           |
| 154601 |     107622 |         13 | 76        |              NULL | NULL           |
| 154602 |     107622 |         15 |           |              NULL | NULL           |
+--------+------------+------------+-----------+-------------------+----------------+
5 rows in set (0.00 sec)
```

glensc:
the queries count is not so good. don't have measurement of previous code, but afaik it had some in memory caching. doctrine is doing some caching by itself, but not all, especially the `matching` type queries.

this is doctrine queries count for single issue:
- `54 statements were executed, 48 of which were duplicates, 6 unique`

non-doctrine queries are accounted separately by debugbar:
- `85 statements were executed, 36 of which were duplicates, 49 unique`

PR #492, opened 2019-03-05
Load/Store user preferences via Doctrine

glensc:
the ui part was split to blocks that are access controlled differently.

preferences are not saved if they are not modified, so changed global default apply immediately
* 6fd94e52e - there is no need to create preference records for new users

also fixes some bugs:
* c5bcd0027 - sendcontroller: handle draft append signature
* e223f58d2 - prevent refresh rate 0 refreshing forever

and relaxes settings save, as they were not verified server side:
* b3b553886 - remove hidden restriction of updating email related settings
* 6abf640bd - relax close_popup_windows setting as well

cc @balsdorf


glensc:
lots of duplication in templates for this same fragment:

```html
<script type="text/javascript">
    <!--
    {if $core.current_user_prefs.close_popup_windows}
        setTimeout('Eventum.close_and_refresh()', 2000);
    {/if}
    //-->
</script>
```

Issue #498, opened 2019-03-09
Getting blank screen on "Associate Emails"


GaneshKandu:
Hi,

Getting blank screen on "Associate Emails" after fresh install of Eventum 3.5.6
Downloaded from [eventum-3.6.2.tar.xz](https://github.com/eventum/eventum/releases/download/v3.6.2/eventum-3.6.2.tar.xz)
[https://github.com/eventum/eventum/releases](https://github.com/eventum/eventum/releases)

ERROR LOG
```
[2019-03-09 00:39:09] app.ERROR: Uncaught Exception TypeError: "Argument 1 passed to Prefs::getUserPreference() must be of the type integer, null given, called in src/Controller/EmailsController.php on line 87" at lib/eventum/class.prefs.php line 24 {"exception":"[object] (TypeError(code: 0): Argument 1 passed to Prefs::getUserPreference() must be of the type integer, null given, called in src/Controller/EmailsController.php on line 87 at lib/eventum/class.prefs.php:24)"} {"url":"htdocs/emails.php","ip":"192.168.1.25","http_method":"GET","server":"mydomain","referrer":"http://mydomain/adv_search.php","file":null,"line":null,"class":null,"function":null,"memory_usage":"2 MB","memory_peak_usage":"2 MB","version":"3.6.2","usr_id":"2"}
```
[eventum.log](https://github.com/eventum/eventum/files/2948114/eventum.log)

PR #499, opened 2019-03-09
Getting blank screen on "Associate Emails"

GaneshKandu:
#498 
```$this->usr_id``` is not defined at line 88 ```src/Controller/EmailsController.php```


glensc:
replaced by #500

PR #506, opened 2019-03-15
Add caching to custom fields

glensc:
Refs:
- #466
- https://tideways.com/profiler/blog/5-doctrine-orm-performance-traps-you-should-avoid


glensc:
unfinished; but merging to have 3.6.3 release

PR #510, opened 2019-03-22
Split extension features to each interface

glensc:
- Deprecates `AbstractExtension` and `ExtensionInterface`
- Breaks compatibility: parameter and return type hints are required now

This allows writing single-purpose extensions:
- an extension providing only `SubscriberProvider` interface

cc @balsdorf


glensc:
Sadly type parameters can't be made forward-compatible:

```php
    public function registerAutoloader(ClassLoader $loader): void
```

i.e for return types, I can make an extending class to use return values, and then update parent class, but with type parameters, there's only one way: either type is there or type is not there.

glensc:
backward incompatible change removed out, so the change can go into 3.6.5

PR #513, opened 2019-03-25
Fix notification role_id type notice

glensc:
Argument 2 passed to [Custom_Field::formatUpdatesToDiffs()](https://github.com/eventum/eventum/blob/v3.6.3/lib/eventum/class.custom_field.php#L113) must be of the type integer or null, string given, called in [lib/eventum/class.notification.php](https://github.com/eventum/eventum/blob/v3.6.3/lib/eventum/class.notification.php#L719) on line 719

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Upgrading to 3.6.x versions requires that you upgrade to latest 3.5.x version fi
 
 - Remove last PEAR dependency: `Math_Stats` (@glensc, #509)
 - Add boot and shutdown events (@glensc, #511)
+- Regression: Fix notification `role_id` type notice (@glensc, #513)
 
 [3.6.4]: https://github.com/eventum/eventum/compare/v3.6.3...master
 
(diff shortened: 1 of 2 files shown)


glensc:
@balsdorf this is a rather interesting problem. appeared once type hints were added to methods (087ddfe2b519853bf0bac9e7dfdcefac876415f2).

The overall problem is that `Notification::getUsersByIssue()` returns `pru_role = ''` for user who created the issue via email, but they are not member of the project (missing relation in `project_user` table).

so this results iterating over such [`$users`](https://github.com/eventum/eventum/blob/v3.6.3/lib/eventum/class.notification.php#L684) array:
```php
Array
(
[0] => Array
    (
        [sub_usr_id] => 1085
        [sub_email] =>
        [pru_role] => 5
    )

[1] => Array
    (
        [sub_usr_id] => 1161
        [sub_email] =>
        [pru_role] =>
    )
)
```

what's the proper behavior here? I could try to figure what's the behavior with previous code where the type was not checked, but I don't think it's accurate behavior.

glensc:
merging to fix 3.6.3 regression, @balsdorf can still feedback after merge.

PR #393, opened 2018-06-27
Security 2018 05

balsdorf:
* [CVE-2018-12628] Add CSRF protection to user and project admin pages (@balsdorf, Yaroslav Babin)
* [CVE-2018-12621] Fixed open redirect (@balsdorf, Yaroslav Babin)
* [CVE-2018-11569] Fixed remote code execution vulnerability (@balsdorf, Yaroslav Babin)
* [CVE-2018-12622, CVE-2018-12623, CVE-2018-12624, CVE-2018-12625, CVE-2018-12626, CVE-2018-12627] XSS fixes (@balsdorf, Yaroslav Babin)

PR #530, opened 2019-04-22
Use array json decode for cst_custom_field

glensc:
https://github.com/eventum/eventum/pull/393 changed `cst_custom_field ` serialization from `serialize` to `json_encode`, but this caused wrong data type, when it was previously array, it's now `stdClass`, and templates can't handle that:

on `adv_search.php?custom_id=1`:

```
[2019-04-22 23:07:37] app.ERROR: Uncaught Exception Error: "Cannot use object of type stdClass as array" at /Users/glen/scm/eventum/eventum/var/cache/5f42ad00d36c4e9912c1dbb8862f873575e473ae_0.file.adv_search.tpl.html.php line 95 {"exception":"[object] (Error(code: 0): Cannot use object of type stdClass as array at /Users/glen/scm/eventum/eventum/var/cache/5f42ad00d36c4e9912c1dbb8862f873575e473ae_0.file.adv_search.tpl.html.php:95)
```

this PR changes seralization to use arrays.

unclear how migration was handled in to different type in #393 so nothing changed here as well.

cc @balsdorf

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 Upgrading to 3.6.x versions requires that you upgrade to latest 3.5.x version first.
 
 - Avoid fatal error if declared workflow class can not be loaded (@glensc, #529)
+- Use array json decode for `cst_custom_field` (@glensc, #530)
 
 [3.6.7]: https://github.com/eventum/eventum/compare/v3.6.6...master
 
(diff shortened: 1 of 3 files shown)


glensc:
this is maybe fix for https://github.com/eventum/eventum/issues/393

cc @areis422

PR #532, opened 2019-04-25
Add open issues count next to assignees in Reports section

vladsf:
Add open issues count next to assignee names (http://eventum.local/reports/open_issues.php). 
HTML template fix. This is a very minor feature.


glensc:
please fill changelog entry

PR #549, opened 2019-05-01
Use ocramius/package-versions for version info

glensc:
use [ocramius/package-versions] package for version info

[ocramius/package-versions]: https://packagist.org/packages/ocramius/package-versions

could also use:
- https://github.com/Jean85/pretty-package-versions
- https://github.com/nikolaposa/version


glensc:
noticed some minor issue with version extraction: https://github.com/Ocramius/PackageVersions/issues/84

PR #562, opened 2019-05-10
GitLab: skip issue matching on empty text

glensc:
in case of issue deletion the description in payload is NULL

`{"code":-1,"message":"Argument 1 passed to Eventum\\IssueMatcher::match() must be of the type string, null given, called in /src/Scm/Adapter/Gitlab.php on line 98"}`

```json
{
  "object_kind": "issue",
  "event_type": "issue",
...
  "object_attributes": {
    "author_id": 68,
    "closed_at": null,
    "confidential": false,
    "created_at": "2019-05-09 15:23:45 +0300",
    "description": null,
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 Upgrading to 3.7.x versions requires that you upgrade to latest 3.5.x version first.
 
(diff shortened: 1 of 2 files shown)


glensc:
unable to reproduce this. so just merging

PR #579, opened 2019-05-29
A null $config['options'] throwing uncaught exception error: "Call to a member function toArray() on null"

basepr1me:
Simple patch to check $config['options'] for null to open Setup Authentication in UI.


glensc:
why is it missing for you? did you skip some upgrade steps? are all migrations applied?

the `options` key is added by this patch:
- [db/migrations/20190115211455_eventum_auth_adapter_setup.php](https://github.com/eventum/eventum/blob/v3.7.1/db/migrations/20190115211455_eventum_auth_adapter_setup.php)

glensc:
also, some git fu:
- use `git pull --rebase` instead of creating merge commits
- `git commit --amend` if you intend to fix last commit

maybe useful reading in this topic:
- https://www.atlassian.com/git/tutorials/syncing/git-pull
- http://gitready.com/advanced/2009/02/11/pull-with-rebase.html
- https://stackoverflow.com/questions/2472254/when-should-i-use-git-pull-rebase

glensc:
also, you should never submit pull request from your master branch. always create branch specific to the changes. as if you add more commits to your `master` branch, they will end up in this pull request.

basepr1me:
Understood on pull requests.

I migrate from release-to-release, running upgrade.php for each release. I'll look further. Thanks.

glensc:
unclear what was happening; this should not happen if patches applied properly.

PR #640, opened 2019-08-08
Update for new EventDispatcherInterface::dispatch() signature

glensc:
https://symfony.com/blog/new-in-symfony-4-3-simpler-event-dispatching


glensc:
[`Psr\EventDispatcher\StoppableEventInterface`](https://github.com/php-fig/event-dispatcher/blob/0.7.0/src/StoppableEventInterface.php) requires `php-fig/event-dispatcher `0.7.0 which [requires php 7.2](https://github.com/php-fig/event-dispatcher/blob/0.7.0/composer.json#L14)

glensc:
`object` return type hint appeared in php 7.2:
- https://wiki.php.net/rfc/object-typehint

PR #646, opened 2019-08-13
Fix scm_ping route

glensc:


glensc:
support to resubmit the failed scm pings: https://github.com/eventum/scm/pull/7

PR #629, opened 2019-08-02
Move constants from config.php to setup.php

glensc:
Global constants are difficult to maintain, and having constants configurable from config file is a super odd concept making them not really constants.

This is to cleanup codebase and move most of the constants to `setup.php` config array.
- [x] `APP_CHARSET`: replaced with `UTF-8` values.
- [x] `APP_DEFAULT_ASSIGNED_EMAILS`
- [x] `APP_DEFAULT_NEW_EMAILS`
- [x] `APP_DEFAULT_COPY_OF_OWN_ACTION`
- [x] `APP_DEFAULT_PAGER_SIZE`
- [x] `APP_DEFAULT_REFRESH_RATE`
- [x] `APP_DEFAULT_TIMEZONE`
- [x] `APP_DEFAULT_WEEKDAY`
- [x] `APP_SITE_NAME`
- [x] `APP_EMAIL_ENCODING`: drop, not needed
- [x] `APP_GETTEXT_MODE`
- [x] `APP_ROUTED_MAILS_SAVEDIR`
- [x] `APP_INTERNAL_COLOR`
- [x] `APP_DEFAULT_LOCALE`
- [x] `APP_ENABLE_FULLTEXT`, `APP_FULLTEXT_SEARCH_CLASS`
- [x] `APP_SYSTEM_USER_ID`
- [x] `APP_SYSTEM_USER_ID`
- [x] `APP_COOKIE`, `APP_COOKIE_EXPIRE`, `APP_PROJECT_COOKIE`, `APP_PROJECT_COOKIE_EXPIRE`
- [x] `APP_COOKIE_DOMAIN`
- [x] `APP_COOKIE_URL`
- [x] `APP_BASE_URL`
- [x] `APP_RELATIVE_URL`
- [x] `APP_ERROR_LOG` drop, usage removed in 70441d866

PR #704, opened 2019-10-21
Drop deprecated config/config.php

glensc:
Followup to https://github.com/eventum/eventum/pull/629: Drop config/config.php support.

Update to latest 3.8.x is required for previous `config.php` constants to be migrated.


glensc:
one more PR that should be released before this can be applied:
- https://github.com/eventum/eventum/pull/811

glensc:
And one more PR that delays this: https://github.com/eventum/eventum/pull/824

PR #797, opened 2020-03-09
LDAP: Support encryption

inguin:
Supported options are: No encryption, LDAP over SSL and LDAP with
StartTLS.


glensc:
Can you please add Changelog entry

inguin:
> Can you please add Changelog entry

Done.

PR #812, opened 2020-04-08
Update jquery chosen from 1.2.0 to 1.8.7

glensc:
Hopefully resolves broken dropdown ui:

image

cc @slay123


glensc:
did not have the intended effect, but merging anyway to get other updates

PR #815, opened 2020-04-18
Switch to symfony/lock implementation

glensc:
Same feature set (we're using flock), except maybe a bit more popular framework in use


glensc:
Also:
- https://symfony.com/blog/new-in-symfony-3-2-console-improvements-part-2
- https://symfony.com/doc/4.4/console/lockable_trait.html

glensc:
The LockableTrait is not configurable:
- uses SemaphoreStore if extension present (we want FlockStore)
- uses sys_get_temp_dir (we want Paths::APP_LOCKS_PATH)

PR #816, opened 2020-04-19
Update league/commonmark to 1.4.0

glensc:
This allows using new extensions:
- Added new Heading Permalink extension (https://github.com/thephpleague/commonmark/pull/420)
- Added new Table of Contents extension (https://github.com/thephpleague/commonmark/issues/441)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,7 @@
 		"horde/text-flowed": "dev-patch-1 as 2.0.3",
 		"horde/util": "dev-patch-1 as 2.5.8",
 		"ircmaxell/random-lib": "^1.2",
-		"league/commonmark": "^1.3",
+		"league/commonmark": "^1.4",
 		"league/flysystem": "^1.0",
 		"league/html-to-markdown": "^4.8",
 		"malkusch/lock": "^2.1",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -44,7 +44,7 @@
 		"horde/text-flowed": "dev-patch-1 as 2.0.3",
 		"horde/util": "dev-patch-1 as 2.5.8",
 		"ircmaxell/random-lib": "^1.2",
-		"league/commonmark": "^1.3",
+		"league/commonmark": "^1.4",
```
GrahamCampbell, line 47: Since you are using 1.4, you should change the code to use the non-deprecated classes.

glensc: @GrahamCampbell thanks for noticing.

This is still WIP (GitHub doesn't allow on/off WIP state, like GitLab and I forgot on initial PR creation)

GrahamCampbell: > GitHub doesn't allow on/off WIP state

It does actually. They launched this a couple of weeks ago. ;)

glensc: @GrahamCampbell: oh.

PR #826, opened 2020-05-05
Add Pimple based service container

glensc:
Add [Pimple] - A simple PHP Dependency Injection Container, as it's what I'm familiar with.

[Pimple]: https://pimple.symfony.com/

Changing later to PSR11 container, for example [Symfony Service Container], is smaller task than refactoring code to use containers and DI in general.

[Symfony Service Container]: https://symfony.com/doc/4.4/service_container.html


glensc:
This is the first iteration and moves ExtensionManager to be used from the container.

PR #833, opened 2020-05-09
Use Doctrine methods to save in ProjectsController

glensc:
Extracted out of https://github.com/eventum/eventum/pull/832


glensc:
Some future refactoring, discard the options serialization, use actual column:

```
mysql> select prj_anonymous_post,prj_anonymous_post_options from project where prj_id=1\G
*************************** 1. row ***************************
        prj_anonymous_post: enabled
prj_anonymous_post_options: a:5:{s:18:"show_custom_fields";s:2:"no";s:8:"reporter";s:1:"3";s:8:"category";s:1:"2";s:8:"priority";s:1:"1";s:5:"users";a:1:{i:0;s:1:"3";}}
1 row in set (0.00 sec)
```

it's also weird that anonymous reporting has different manage screen (manage/anonymous.php?prj_id=1), but saved to main project table.

PR #837, opened 2020-05-10
Extract ExpandableCell into a class

glensc:
Load ExpandableCell via webpack. This serves as an example how to define javascript modules.


glensc:
merging, to see if this minimal change causes any issues, like browser compatibility.

PR #850, opened 2020-05-24
Migrate Email_Account class logic to Models and Doctrine Repository

glensc:
This is code cleanup that makes use of current usage patterns.


glensc:
Changeset too big will continue in separate MR updating rest of the logic.

PR #857, opened 2020-05-28
Fix privilege check when closing issues

inguin:
The privilege check in CloseController is insufficient, it does not
verify that the user actually has permissions to update the issue. A
user with 'Viewer' or 'Reporter' role can close issues by entering the
close.php URL.


glensc:
thanks!

PR #873, opened 2020-06-26
Switch to composer 2.0 when making release tarball

glensc:
- [x] Enable composer 2.0 platform check: https://github.com/composer/composer/blob/2.0.0-alpha2/doc/07-runtime.md#platform-check
- [ ] Use composer-runtime-api 2.0


glensc:
Can't just easily switch to 2.0:

```
[~/scm/eventum/eventum (composer-platform-check)↕★] ➔ ./composer.phar self-update --preview
[~/scm/eventum/eventum (composer-platform-check)↕★] ➔ ./composer.phar install
The "ocramius/package-versions" plugin was skipped because it requires a Plugin API version ("^1.0.0") that does not match your Composer installation ("2.0.0"). You may need to run composer update with the "--no-plugins" option.
The "hirak/prestissimo" plugin was skipped because it requires a Plugin API version ("^1.0.0") that does not match your Composer installation ("2.0.0"). You may need to run composer update with the "--no-plugins" option.
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Your lock file does not contain a compatible set of packages. Please run composer update.

  Problem 1
    - ocramius/package-versions is locked to version 1.4.2 and an update of this package was not requested.
    - ocramius/package-versions 1.4.2 requires composer-plugin-api ^1.0.0 -> found composer-plugin-api[2.0.0] but it does not match the constraint.
  Problem 2
    - ocramius/package-versions 1.4.2 requires composer-plugin-api ^1.0.0 -> found composer-plugin-api[2.0.0] but it does not match the constraint.
    - doctrine/orm v2.7.3 requires ocramius/package-versions ^1.2 -> satisfiable by ocramius/package-versions[1.4.2].
    - doctrine/orm is locked to version v2.7.3 and an update of this package was not requested.

Use the option --with-all-dependencies to allow upgrades, downgrades and removals for packages currently locked to specific versions.
```

glensc:
This is outdated and no longer. besides `symfony/flex` puts version check to vendor/autoload.php, which was the initial intent to enable platform check.

```php
# cat vendor/autoload.php
<?php

if (\PHP_VERSION_ID < 70300) {
    echo sprintf("Fatal Error: composer.lock was created for PHP version 7.3.0 or higher but the current PHP version is %d.%d.%d.\n", PHP_MAJOR_VERSION, PHP_MINOR_VERSION, PHP_RELEASE_VERSION);
    exit(1);
}

// autoload.php @generated by Composer

require_once __DIR__ . '/composer/autoload_real.php';

return ComposerAutoloaderInitEventumCore::getLoader();
```

PR #880, opened 2020-06-29
Show user inactive status

glensc:
Shows user status if it is not "active":

image

image

- [x] reporter in issue view
- [x] edit notification popup
- [x] notes listing


slay123:
> @slay123 please indicate more places where you'd like to see the status

Hey, I would say that in Notification list, but Inactive members aren't shown there, although if you click on Notification List, then they are listed there, in "Existing Subscribers for this Issue ". So perhaps there?

PR #883, opened 2020-06-30
Fix getPassword to return EncryptedValue object

glensc:
Fixes regression from #850 that encryption can't be enabled without an error.


glensc:
maybe could use doctrine encrypted type for this in the future
- https://stackoverflow.com/questions/44634706/doctrine-encrypt-column
- https://github.com/somecoding/laminas-doctrine-encrypt

PR #890, opened 2020-07-16
Move SCM setup to Extension

glensc:
If the extension is not enabled, the `/scm_ping` route will be missing.


glensc:
Tried to configure /scm_ping route from extension, but it's complicated because routes config is cached based on resources.

- https://symfony.com/doc/5.1/components/config/caching.html

glensc:
Having routes cached is okay, it's new reality that config is cached and if you make changes cache clea is needed.

For example, the same problem appears with this PR:
- https://github.com/eventum/eventum/pull/1024

For both solutions, probably need to add cache warmup after config changes for production mode (for dev mode it's transparent). The cache warmup was already done in setup as webbug pixel:
- https://github.com/eventum/eventum/pull/1007

Issue #936, opened 2020-09-15
Q: remove php-imap requirement


vladsf:
Hi,

php-imap package removed from RHEL8 and derivatives.
Getting PHP 7.2 with php-imap extension is not out of box anymore.
php-imap depends on a pretty old libc-client which in turn requires openssl 1.0 which makes things troublesome in the near future.
https://bugzilla.redhat.com/show_bug.cgi?id=1755408

Is there any plan to workaround obsoleting php-imap and libc-client?

glensc:
no actual plan. switching to zend\mail was so overwhelming that switching pop3/imap part was just forgotten.

i think pieces are even in place to use zend\mail for this.

glensc:
some links:
- https://github.com/glpi-project/glpi/issues/3481

glensc:
actually the alternative planned to use zend (now laminas) components, but they plan to put the mail component to security-only:
- https://github.com/laminas/technical-steering-committee/pull/31

and so far I have not found any alternative to parse input emails.

we migrated from pear/mail, which was huge project, took several iterations to complete (3.4.0 2018-04-22):
- Drop `pear/mime_decode`, `pear/mail_mime` in favor of `Zend\Mail` (@glensc, #263, #333, #334, #335)

vladsf:
I guess moving from an old php-imap used by many  to obsoleting laminas component is not a good idea. php-imap [gets patches](https://github.com/uw-imap/imap/tree/cab109466534e206a3652ef1c68fe88101b68bda) and many face the same question.
I can see it gets openssl 1.1 patch and thus it is not abandoned and I could be wrong in my initial post.

PR #937, opened 2020-09-16
Move imap related methods to ImapConnection class

glensc:
Move expunge emails logic to own class.

A small step for moving `ext-imap` code to single class.

Imap related methods are only in `ImapConnection` and `ImapResource` classes.

refs:
- https://github.com/eventum/eventum/issues/936


glensc:
I'm not really able to test the expunge functionality, as I don't have a test IMAP server. So I'll merge as is.

Issue #939, opened 2020-09-18
Wrong config file name in the documentation?


federico-razzoli:
Please check:

https://github.com/eventum/eventum/blob/master/docs/wiki/System-Admin/Displaying-PHP-errors.md

The mentioned file `config/config.php` does not exist on a fresh install. If I should create the file myself, the documentation should mention this in a place that is easy to find. If it means `config/setup.php` or another file it should be fixed.

glensc:
that's all outdated, submitted update: https://github.com/eventum/eventum/pull/940

PR #940, opened 2020-09-18
Update documentation about error logs

glensc:
Fixes https://github.com/eventum/eventum/issues/939


glensc:
if some authentic link exists to guide user more in this area I'd add the link, but at this point, I haven'tt have looked for such page.

PR #947, opened 2020-10-01
Fix sorting links when searching for custom fields

inguin:
Commit 5bb5631be changed the serialization of custom field filter
parameters from PHP array syntax to JSON syntax, but missed one
occurrence in the function building the URLs for sorting links. As a
result custom field filters are lost when changing the sort order by
clicking one of the list headings.

x-ref:
- https://github.com/eventum/eventum/pull/393
- https://github.com/eventum/eventum/pull/530

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 See [Upgrading] for details how to upgrade.
 
+- Fix sorting links when searching for custom fields, #947
+
 [3.9.7]: https://github.com/eventum/eventum/compare/v3.9.6...master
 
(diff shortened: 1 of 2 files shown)


glensc:
please see if it should return array, like fix in https://github.com/eventum/eventum/pull/530

inguin:
#530 only changed `json_decode()` calls, this is `json_encode()` which doesn't have an `$assoc` parameter.

This changes makes the sorting link URLs consistent with the saved search URLs (see line 171).

glensc:
Ah, thanks for looking into that.

PR #960, opened 2020-11-01
Add eventum upgrade matrix to documentation

glensc:


glensc:
@vladsf what do you think of these instruction updates?

docs/wiki/Upgrading.md
```diff
@@ -1,24 +1,51 @@
...
+# Updating Eventum versions
 
 One of our objectives is to make upgrading from an earlier release as
 painless as possible, and we provide scripts that should bring your
 existing Eventum installation up-to-date.
 
+Eventum supports sequential upgrade path using released minor versions.
+
+In other words, upgrade to minor version separately, you can skip patch
+versions in minor versions.
```
vladsf, line 10: I would say like this: "To explain more clearly one may skip all patch versions and upgrade to the next minor version."

docs/wiki/Upgrading.md
```diff
@@ -1,24 +1,51 @@
...
 See 2.2 upgrade instructions how to convert database to UTF-8.
 
-## IMPORTANT
+## Upgrade Process
 
 When upgrading to a new version of Eventum, please follow these instructions:
 
 1.  Check the [requirements](Prerequisites.md) for the version
 1.  Backup your copy of Eventum - files and data.
-2.  Extract your new Eventum copy over your existing folder structure
```
vladsf, line 20: This line was actually helpful: it clearly states that Eventum path matters and the new Eventum shall reside in the same path so config.php has all right paths.

docs/wiki/Upgrading.md
```diff
@@ -1,24 +1,51 @@
-# Upgrade Process
+# Updating Eventum versions
```
vladsf, line 1: may be: Upgrade steps or Upgrade guidelines

docs/wiki/Upgrading.md
```diff
@@ -1,24 +1,51 @@
...
+
 Please note that if your database encoding is not UTF-8, you may encounter various bugs:
 
 -   PDO Driver is [unusable](https://github.com/eventum/eventum/pull/167)
 -   Issue History entries [get corrupted](https://gitter.im/eventum/eventum?at=58225f1d45c9e3eb4314b58c) (JSON requires UTF-8 encoding)
 
 See 2.2 upgrade instructions how to convert database to UTF-8.
 
-## IMPORTANT
+## Upgrade Process
```
vladsf, line 39: IMPORTANT was really good.

docs/wiki/Upgrading.md
glensc: But using this overwrite step, removed files from the new version do not get removed, they stay lingering around...

docs/wiki/Upgrading.md
glensc: 454f4f39a

docs/wiki/Upgrading.md
glensc: this is page heading. there's such subsection already: "Upgrade Process", altho there's other section that imho should be renamed: "Upgrading to latest version"

docs/wiki/Upgrading.md
glensc: but it made no sense, it was just an overview of the process. everything on the page is important, otherwise, if it's no important I rather drop the noise :)

docs/wiki/Upgrading.md
vladsf: Yes, you are right. That's not good. My idea was that to unpack the new Eventum release into the same path. OK, let it be removed.

docs/wiki/Upgrading.md
vladsf: May be just "Upgrading Eventum"?

docs/wiki/Upgrading.md
vladsf: Ok, may be "Upgrade path"? "Upgrade process" sounds to me like a real running process, command line tool.

glensc: and "path" sounds to me like a directory, changing to "overview" then

docs/wiki/Upgrading.md
glensc: That detail is included now, that if you use new installation path, you need to adjust configs.

docs/wiki/Upgrading.md
```diff
@@ -1,37 +1,67 @@
...
-1.  Backup your copy of Eventum - files and data.
-2.  Extract your new Eventum copy over your existing folder structure
-3.  Run the upgrade scripts described in below
+1.  Backup your copy of Eventum - files and data
+1.  Rename the current Eventum folder out of the way (`eventum.old`)
+1.  Extract new Eventum version to existing installation path (`eventum`)
+1.  Restore config and workflow files from the previous version
+1.  Run the upgrade script
+
+This way of installing will get rid of files that got removed from newer Eventum version.
```
vladsf, line 50: It is hard to read. This might sound better: "This way of installing clears all obsolete files from Eventum installation path".

docs/wiki/Upgrading.md
```diff
@@ -1,37 +1,67 @@
...
 When upgrading to a new version of Eventum, please follow these instructions:
 
 1.  Check the [requirements](Prerequisites.md) for the version
-1.  Backup your copy of Eventum - files and data.
-2.  Extract your new Eventum copy over your existing folder structure
-3.  Run the upgrade scripts described in below
+1.  Backup your copy of Eventum - files and data
+1.  Rename the current Eventum folder out of the way (`eventum.old`)
+1.  Extract new Eventum version to existing installation path (`eventum`)
+1.  Restore config and workflow files from the previous version
```
vladsf, line 47: Older statement was clear: "Copy all config files from old version to new version: `eventum.old/config` to `eventum/config`" This time it is not obvious which files to copy.

vladsf (commented):
I'm afk till next week. May be late with responses. Changes looks good to me.

docs/wiki/Upgrading.md
glensc: It is still there, please read the whole documentation.

docs/wiki/Upgrading.md
glensc: Why it's hard to read? People don't understand such complicated words as "obsolete" and apparently don't bother to consult a dictionary either.

Here's a real-life example:
- https://github.com/request/request/issues/3142

PR #961, opened 2020-11-01
Use GitHub Actions for CI integration

glensc:
Travis integration has become totally unbearable recently:
- jobs wait queue is very long, some hours or even days
- status updates are no longer updated in PRs
- inconsistent and old composer versions, making it very difficult to have a consistent environment for different PHP versions. for example, unwanted upgrade to v2 and force to use v1 requires hack as `--1` flag not supported on composer 1.8

🙇 Thank you Travis-CI for being out there for that long, but it is time to switch...

This migrates CI jobs for pull requests to GitHub Actions, the release step is still run in Travis CI.

Inspired by my previous work:
- https://github.com/perftools/xhgui/pull/366

refs:
- https://github.com/actions/toolkit/blob/main/docs/commands.md#environment-files

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,6 +38,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php }}
+          tools: composer:v1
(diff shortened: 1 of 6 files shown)


glensc:
683eb5f before squash

PR #966, opened 2020-11-16
Move EventDispatcher setup to EventDispatcherService service provider

glensc:
The intent is to deprecate EventManager class


glensc:
This code created dependency loop:

```
app.ERROR: Unable to create Subscriber\IrcNotifySubscriber: Maximum function nesting level of '250' reached, aborting! {"exception":"[object] (Error(co
de: 0): Maximum function nesting level of '250' reached, aborting! at src/ServiceContainer.php:29)
```

glensc:
Recursion test:

```
[~/scm/eventum/eventum (EventDispatcherService)★] ➔ ./bin/console.php -vvv
Segmentation fault: 11
```

Issue #963, opened 2020-11-03
Eventum keeps saying "Please select a valid date of work"


eventumrcf:
**Version(s) affected**: 3.9.6
**PHP Version**: 7.3
**Browser and Version**: Chrome 86.0.4240.111
**Description**  

Hi I upgraded to the latest version, but now am experiencing issues:

1) Whenever I try to add a time tracking entry I get the message "Please select a valid date of work" (SOLVED)

2) When downloading e-mail I get 

app.ERROR: Eventum\Model\Repository\IssueRepository 0 not found

Any insights would be appreciated.
Ruan

eventumrcf:
Hi managed to sort it out

glensc:
Please explain how for future readers.

martin-dufka:
How it was "solved"? I have the same error "Please select a valid date of work" in 3.9.9 after update.

PR #984, opened 2021-01-12
Time tracking - Eventum still says "Please select a valid date of work"

martin-dufka:
#963 
- https://github.com/eventum/eventum/blob/master/res/assets/scripts/Validation.js#L146-L156
```js
    isValidDate(field_prefix) {
        const selected_date = new Date();

        selected_date.setMonth(Eventum.getField(field_prefix + '[Month]').val() - 1);
        selected_date.setDate(Eventum.getField(field_prefix + '[Day]').val());
        selected_date.setYear(Eventum.getField(field_prefix + '[Year]').val());

        return selected_date.getDate() === Eventum.getField(field_prefix + '[Day]').val();
    };
```
typeof(selected_date.getDate() ) is number
typeof(Eventum.getField(field_prefix + '[Day]').val()) is string

Function isValidDate is never true in strict comparing. Solution may be retype getField().val() to number and rebuild app.js, I think.

### DIFF
--- a/res/assets/scripts/Validation.js
+++ b/res/assets/scripts/Validation.js
@@ -152,7 +152,7 @@ export class Validation {
         selected_date.setDate(Eventum.getField(field_prefix + '[Day]').val());
         selected_date.setYear(Eventum.getField(field_prefix + '[Year]').val());
 
-        return selected_date.getDate() === Eventum.getField(field_prefix + '[Day]').val();
+        return selected_date.getDate() === Number(Eventum.getField(field_prefix + '[Day]').val());
     };
 
     checkFormSubmission(form, callback_func) {


glensc:
Indeed, b4290376e8 changed the comparison to be strict. So this is broken since 3.8.13

martin-dufka:
I updated from 3.8.3. :)
The permanent link is modified, thanks for the file link warning. I'm still learning... :)

glensc:
you can also just use the actual tag instead of commit. that's what I usually do as I'm pedant :D, even just using `y` shortcut is way easier.

martin-dufka:
It's changed to tag now. :)

martin-dufka:
commit auto-includes source code, tag doesn't. So the last edit, I hope.... :) :) :)

PR #986, opened 2021-01-22
Fix decoding of non-plaintext parts in multipart mails

inguin:
Parts of multipart MIME messages are often sent in quoted-printable
or base64 encoding. Use the DecodePart helper to decode them based on
their Content-Transfer-Encoding header. This was already done for
text/plain parts, but not for others like text/html.

### DIFF
--- a/src/Mail/Helper/TextMessage.php
+++ b/src/Mail/Helper/TextMessage.php
@@ -88,6 +88,7 @@ class TextMessage
         $filename = $hasDisposition ? $part->getHeaderField('Content-Disposition', 'filename') : null;
         $is_attachment = $disposition === 'attachment' || $filename;
         $charset = $hasContentType ? $part->getHeaderField('Content-Type', 'charset') : null;
+        $content = (new DecodePart($part))->decode();
 
         switch ($contentType) {
             case 'multipart/related':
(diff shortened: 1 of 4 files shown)


glensc:
Can you please add some test-case with raw email as input

src/Mail/Helper/TextMessage.php
```diff
@@ -88,6 +88,7 @@ private function processPart($part): void
         $filename = $hasDisposition ? $part->getHeaderField('Content-Disposition', 'filename') : null;
         $is_attachment = $disposition === 'attachment' || $filename;
         $charset = $hasContentType ? $part->getHeaderField('Content-Type', 'charset') : null;
+        $content = (new DecodePart($part))->decode();
```
glensc, line 91: is this safe to do? i.e decode unconditionally without checking type? I'm afraid the `(new DecodePart($part))->decode()` will either throw exception or produce garbage, for example for application/octet-steream of image/jpeg...

inguin: `decode()` will return the original content unless there is a `Content-Transfer-Encoding` header set to either `quoted-printable` or `base64`. That should work independent of the part type. No reason why an image cannot be base64-encoded.

Currently `decode()` will throw an exception for unsupported encodings. That might be a problem, but not a new one.

inguin:
> Can you please add some test-case with raw email as input

How can I execute these test cases?

glensc:
On 22.01.2021 12:59, Ingo van Lil wrote:

>     Can you please add some test-case with raw email as input
>
> How can I execute these test cases?
>

./vendor/bin/phpunit tests/PathToTest.php --filter testNameOfTestFunction


or use IDE, run test from function name from PHPStorm works.

inguin:
Hope this is what you had in mind. I added an example message containing four HTML parts with different encodings. I can split this into four separate test cases if you like.

PR #860, opened 2020-06-03
Markdown preview and help link

inguin:
Implement Markdown preview and add a help link for the Markdown syntax
to the following text areas:

- New/update issue description
- Internal note
- Bulk close message
- Phone call description
- Manage news
- Manage internal FAQ
- Manage canned email responses

Notes:
- This is pretty basic and has room for improvement, but it works pretty well.
- I'm unsure about the (mis)use of `RemoteDataController` which is otherwise only used for expandable cells.

PR #989, opened 2021-02-25
Add md_textarea.tpl.html include to handle markdown enabled text areas

glensc:
Removes duplication added in https://github.com/eventum/eventum/pull/860

Refs:
- https://github.com/eventum/eventum/pull/860#issuecomment-642548807

### DIFF
--- a/templates/include/md_textarea.tpl.html
+++ b/templates/include/md_textarea.tpl.html
@@ -24,4 +24,7 @@ Parameters:
   tabindex="{$tabindex|default:''}"
   style="{$style|default:''}"
 >{$content|escape:"html"}</textarea>
+
+{include file="error_icon.tpl.html" field="$id"}
+
 <div id="preview" style="width: 97%"></div>
--- a/templates/include/md_textarea.tpl.html
+++ b/templates/include/md_textarea.tpl.html
@@ -1,0 +1,27 @@
+{*
+Parameters:
+  $content: The initial value for textarea
+  $issue_id: The issue id
+  $id: The id attribute to textarea
+  $name: The name attribute to textarea
+  $rows: The rows attribute to textarea
+  $cols: The cols attribute to textarea
+  $tabindex: The tabindex attribute to textarea
+  $style: The style attribute to textarea
+*}
+
+<a class="white_link" href="javascript:void(null);"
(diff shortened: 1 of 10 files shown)


templates/include/md_textarea.tpl.html
```diff
@@ -0,0 +1,30 @@
...
+<a class="white_link" href="javascript:void(null);"
+   onClick="Eventum.showPreview('#{$id}', '#preview', { issue_id: {$issue_id|intval|default:''} })">{t}[Preview]{/t}</a>
+<a class="white_link" href="https://commonmark.org/help/" target="_blank">{t}[Markdown Help]{/t}</a>
+<br>
+<textarea
+  id="{$id}"
+  name="{$name|default:$id}"
+  rows="{$rows|default:''}"
+  cols="{$cols|default:''}"
+  tabindex="{$tabindex|default:''}"
```
glensc, line 24: i wonder does html like `cols="" rows=""` or `tabindex=""` cause issues in some browser.

works fine for chrome

templates/include/md_textarea.tpl.html
```diff
@@ -0,0 +1,27 @@
...
+<br>
+<textarea
+  id="{$id}"
+  name="{$name}"
+  rows="{$rows|default:''}"
+  cols="{$cols|default:''}"
+  tabindex="{$tabindex|default:''}"
+  style="{$style|default:''}"
+>{$content|escape:"html"}</textarea>
+<div id="preview" style="width: 97%"></div>
```
inguin, line 30: Maybe use a unique ID like `${id}-preview` to avoid conflicts in case of multiple markdown text areas on one page. The calls to `showPreview()` and `hidePreview()` above would need to be adapted as well.

templates/include/md_textarea.tpl.html
inguin: No issues in Firefox either.

templates/include/md_textarea.tpl.html
glensc: I omitted this deliberately as:
1. it works identically to previous code.
1. this is going to go away with rewrite to prosemirror PR I will submit next

glensc: https://github.com/eventum/eventum/pull/991

PR #991, opened 2021-02-26
Add initial setup of prosemirror-markdown for markdown

glensc:
Add initial setup of prosemirror-markdown for markdown textareas using example implementation:
- https://prosemirror.net/examples/markdown/

This is part of a larger issue https://github.com/eventum/eventum/issues/1023

To use template from this PR modify:

```diff
diff --git a/templates/include/textarea.tpl.html b/templates/include/textarea.tpl.html
index aacfede0e..77c556b12 100644
--- a/templates/include/textarea.tpl.html
+++ b/templates/include/textarea.tpl.html
@@ -1 +1 @@
-{include file="include/textarea_legacy.tpl.html"}
+{include file="include/textarea_prosemirror.tpl.html"}
```


glensc:
The code to implement this came out surprisingly simple!

glensc:
Issues found so far:

- [ ] Editor background should stay the same as textarea
- [ ] editor toolbar needs to be styled, currently white background looks weird
- [ ] Redesign of the switcher needed, it seems 
- [ ] Adjust textarea minimum height, currently it jumps to the content height which on the short body is jumping area 40 rows to 10 rows. and page jumping is always bad.
- [ ] Add link and image buttons popup appears bottom of the page and has no styling. perhaps just disable the buttons for now.
- [ ] tasklist is messed up, but check this solution: https://discuss.prosemirror.net/t/todo-list-improvements/1449/7

Also:
- the previous preview functionality? drop it? it may render differently as different renderer used.

glensc:
- [ ] "description" area validation fails, it thinks the field is empty

switching to WYSIWYM mode and doing some edits can workaround for now.

glensc:
- [ ] tabIndex wrong, shift+tab used. to switch from "textarea" to "subject" input

glensc:
- [ ] the [form autosave](https://github.com/eventum/eventum/pull/145) seems to be non-functioning. staying with some old save

glensc:
- [ ] critical: all notes are submitted as "whatever was saved from garlic"

so, don't use this branch for something you need to submit.

glensc:
this looks like bigger work than I anticipated, help needed.

@inguin, @vladsf if you want to jump into this PR goal, I've updated my fork default branch to be this branch, so just submit your changes to my fork and I'll merge & rebase to include them in this branch.

- https://github.com/glensc/eventum

inguin:
@glensc I had a look at your branch, that looks very promising! I'm rubbish with regard to anything involving HTML styles, so I won't be much help for about half of the issues. I'll try to have a look at things like validation and saving, though.

Ideally the ProseMirror view should be sufficient to fully replace the existing preview feature, provided we can use the same renderer and style.

glensc:
@inguin but we currently render on the server side, you suggest to move that rendering also to client-side?

inguin:
> @inguin but we currently render on the server side, ...

Oh, good point. Also the ProseMirror view probably won't be able to render certain things like issue references or user-defined link filters. So maybe we should keep the preview.

glensc:
@inguin not all of the problems are CSS. I've summarized them to the PR body now. Separated Critical vs Nice to have.

glensc:
I've moved the tasklist to a separate issue:
- https://github.com/eventum/eventum/issues/1023

I'll make this option a feature toggle (default off), to be able to merge existing work.

glensc:
Merging as I'd hate to lost this work.

But as it's not functional, the feature is disabled. See the parent issue for more details and future progress.

PR #1025, opened 2021-03-25
Add extension support to be able to configure container

glensc:
This allows extensions to modify container to add parameters and own services.

```php
class ExampleExtension implements
    Provider\ContainerConfiguratorProvider,
    Provider\ConfigureContainerProvider
{
    /**
     * https://symfony.com/doc/4.4/components/dependency_injection.html
     */
    public function containerConfigurator(ContainerConfigurator $configurator): void
    {
        $services = $configurator->services();
    }

    public function configureContainer(ContainerBuilder $container, LoaderInterface $loader): void
    {
        $loader->load(__DIR__ . '/config/{services}.yml', 'glob');
    }
```

PR #1026, opened 2021-03-25
Refactor ExtensionManager to take extensions list from constructor

glensc:
Addendum to https://github.com/eventum/eventum/pull/1025.

This ensures `ExtensionManager` is final and testable. Somehow #1025 CI passed and merged prematurely event tests were broken.

This attempts to really fix the cyclic loop that #1025 introduced with a small tradeoff of backward-incompatible change.

# Backward incompatible change

Extensions should not access things from container in their constructor. Such access must be delayed, for example use `LazyPropertiesTrait`: fcb1fbd6ae9210312c7e9e1c71f03ecae9c7415c

### DIFF
--- a/init.php
+++ b/init.php
@@ -12,7 +12,6 @@
  */
 
 use Eventum\Event\SystemEvents;
-use Eventum\Extension\ExtensionManager;
 use Eventum\ServiceContainer;
 
 require_once __DIR__ . '/autoload.php';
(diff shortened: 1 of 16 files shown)


glensc:
Yeah, the error was actually a warning:
- https://github.com/eventum/eventum/runs/2191575441

glensc:
And the dependency loop wasn't actually fixed in #https://github.com/eventum/eventum/pull/1025, but avoided rather due to additional bug: `$extensions = []`

PR #1062, opened 2021-04-14
Add HelperTrait to register all lazy properties

glensc:
This replaces one magic with another magic, which is maybe less magical, but still magical.

This allows controllers to define lazy properties without conflicting with base lazy properties:

```php
private $lazyProperties = ['fieldname', 'fieldname2'];
```


glensc:
Some code is missing from this branch. Hmm. I forgot to push?

PR #1063, opened 2021-04-15
Add ISSUE_CREATE_PARAMS event when creating new issue

glensc:
Creating PR of some old changeset (30 Jul 2020).


glensc:
Since this was up in my production for a while, it works by not breaking things, so can be merged.

PR #1200, opened 2021-09-16
Hide the search input on single selects if there are 10 or fewer options

vladsf:
Hide the search input on single selects if there are 10 or fewer options.


glensc:
@vladsf please fill changelog entry

PR #849, opened 2020-05-24
Allow choosing Auth adapters from Extension provides

glensc:
Allows an extension to provide class names that implement `Auth\Adapter\AdapterInterface`.

```php
<?php

namespace Example;

use Eventum\Extension\ClassLoader;
use Eventum\Extension\Provider;
use Example\Auth\NullAuthAdapter;

/**
 * Example Eventum Extension.
 *
 * To enable this extension, write to config/setup.php:
 *
 * 'extensions' => [
 *   'Eventum\\Extension\\ExampleExtension' => '/path/to/this/file/ExampleExtension.php',
 * ],
 */
class ExampleExtension implements Provider\AuthAdapterProvider
{
    /**
     * Return class names implementing Auth\Adapter\AdapterInterface
     *
     * @return string[]
     * @see \Eventum\Extension\Provider\AuthAdapterProvider
     */
    public function getAvailableAuthAdapters(): array
    {
        return [
            NullAuthAdapter::class,
        ];
    }
}

```

Refs:
- https://github.com/eventum/eventum/issues/847

Issue #847, opened 2020-05-22
Q: how to migrate from APP_AUTH_BACKEND?


vladsf:
Hi,
I have Eventum 3.5.5 and custom SSO auth class in eventum/auth/class.my_sso_auth_backend.php
plugged with define('APP_AUTH_BACKEND', 'My_Sso_Auth_Backend');
(Eventum users are synced with a backend script and a web server SSO sets REMOTE_USER)

I can not find APP_AUTH_BACKEND in the latest release. How do I plug my custom SSO in the latest Eventum?

glensc:
This was changed in https://github.com/eventum/eventum/pull/441

There seems to be admin interface to configure adapters:
- `/manage/auth.php`

but it supports only built in versions....

however, if you update your class to implement new [AdapterInterface](https://github.com/eventum/eventum/tree/v3.8.12/src/Auth/Adapter/AdapterInterface.php) you can just load it manually from config/setup.php:

```php
    'auth' => [
        'adapter' => 'Eventum\\Auth\\Adapter\\My_Sso_Auth_Backend',
        'options' => [
        ],
    ],
```

I guess best would be to add support to [extensions](https://github.com/eventum/eventum/tree/v3.8.12/docs/examples/extension) to be able to provide classes for auth providers, then you can configure it using admin interface.

glensc:
I've added such support to select custom auth adapter if your extension provides it:
- https://github.com/eventum/eventum/pull/849

vladsf:
Thanks!

PR #1202, opened 2021-09-20
Add a plug for custom authentication adapter

vladsf:
Hi,

Eventum provides several auth backends, but it has no a plug for a custom one. It used be defined by  'APP_AUTH_BACKEND' but not anymore. This minor patch provides a place for a custom adapter.

The important change is the line in Factory.php

### DIFF
--- a/src/Auth/Adapter/CustomAdapter.php
+++ b/src/Auth/Adapter/CustomAdapter.php
@@ -1,0 +1,108 @@
+<?php
+
+/*
+ * This file is part of the Eventum (Issue Tracking System) package.
+ *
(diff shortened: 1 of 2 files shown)


glensc:
Use extension instead:
- https://github.com/eventum/eventum/pull/849

glensc:
See the same answer given to you last time:
- https://github.com/eventum/eventum/issues/847

vladsf:
I tried #849 but I get `app.ERROR: Unable to load Eventum\Extension\ExampleExtension` even with provided examples.

PR #1203, opened 2021-09-20
Use correct namespace and path for the extension example

vladsf:
`ExampleExtension` class is defined in `Example` namespace so fix the namespaced reference and the file path hardcoded in the 
extension loader function.


glensc:
@vladsf can you try to improve title/body to avoid the click-bait effect. unclear what the change is until I click for the diffs. and probably a description of the bug, and/or link to it.

PR #1206, opened 2021-09-22
Remove hard coded extension path from Example extension

vladsf:
Current Example extension path is almost hard coded in the autoloader function. 
Moving ExampleExtension outside docs/ causes hard to understand errors.


glensc:
1. the code does not compile, you reference a constant that doesn't exist
2. the example is deliberately using the same structure as `vendor/composer/autoload_*.php` files, so you can copy paste values from there.

glensc:
EDIT: looks like the constant was added already earlier:
- https://github.com/eventum/eventum/commit/9d71cd3f5ad357182296c859bd1a1b27da94d718

PR #1208, opened 2021-09-23
Make "Inactive DN" parameter optional as the Help page says

vladsf:
Make "Inactive DN" parameter optional as the [Help](https://github.com/eventum/eventum/blob/a149ed43f345c17570ce3aadeb4d819b81333d10/docs/help/ldap.md) page says. 

Eventum hides "Customize LDAP Config" page if the parameter is not set. `bin/console.php eventum:ldap:sync` fails as well.


glensc:
@vladsf please add changelog entry

glensc:
Hah, I didn't realize @vladsf you have enough mana to merge :D

vladsf:
So it the mana was not added by mistake. OK! :D

glensc:
I guess it was added years ago, and I already forgot

PR #1209, opened 2021-09-24
Add loose email validation in LdapSyncCommand

vladsf:
Skip LDAP user entries with no single valid email. This patch adds a loose check there is at least one valid email.

### DIFF
--- a/src/Console/Command/LdapSyncCommand.php
+++ b/src/Console/Command/LdapSyncCommand.php
@@ -147,8 +147,9 @@ class LdapSyncCommand extends BaseCommand
     {
         $users = $this->ldap->getUserListing($dn);
         foreach ($users as $user) {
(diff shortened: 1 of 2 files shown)


src/Console/Command/LdapSyncCommand.php
```diff
@@ -158,6 +159,23 @@ private function findUsers($dn): Generator
         }
     }
 
+    /**
+     * Find a valid email
+     *
+     * @param string[] $emails
+     * @return string|null
+     */
+    private function findValidEmail(array $emails): ?string
```
glensc, line 168: you're not using the value, so rename the method to `hasValidEmail(array $emails): bool`

glensc:
sync pull request title with changelog entry. changelog entry is more precise

PR #1210, opened 2021-09-24
Make user notification on account creation optional

vladsf:
Do not send email notice if user password is not set. 
If user accounts are created outside Eventum e.g. in LDAP and then ldapsync'ed then there is no reason to send email notice with
empty password.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ See [Upgrading] for details on how to upgrade.
 
 - Hide the search input on single selects if there are 10 or fewer options, #1200
 - Make LDAP "Inactive DN" parameter optional, #1208
+- Make user notification on account creation optional, #1210
 
 ## [3.10.6] - 2021-08-03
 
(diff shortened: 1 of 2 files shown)


glensc:
if my users are external none of them have passwords set in the eventum database.

vladsf:
OK, I made notification email optional on user creation. The same way [User::update()](https://github.com/eventum/eventum/blob/76a6f6dc60100d3fa74d43f95b82a7b64a1005b0/src/Auth/Adapter/LdapAdapter.php#L279) does. 

This change does not affect current workflow and only repeats `User::update()` call.

This allows to call User::insert() with disabled notification from [LdapAuthAdapter.php](https://github.com/eventum/eventum/blob/76a6f6dc60100d3fa74d43f95b82a7b64a1005b0/src/Auth/Adapter/LdapAdapter.php) later.

glensc:
please rebase against origin/master, rather create merge commit in merge request.

but I'll squash merge this time. consider this if you want to preserve your original commit messages after merge.

vladsf:
Ack. Thanks!

PR #1216, opened 2021-09-27
Add example command for extension

vladsf:
Here is an example command I trying to add to a release of Eventum, but I get 
the error:

```
[eventumtest Command]$ /eventum/bin/console.php eventum:example:hello
  There are no commands defined in the "eventum:example" namespace.
..
```
Not sure I understand how to add custom console commands.


glensc:
In other words, you need to clear the cache for the dependency injection:
- https://symfony.com/doc/4.4/console.html#registering-the-command

ps: cross-link this PR and your discussion for other people!

glensc:
I've refactored the class to move it to extension, so this can be merged.

glensc:
@vladsf by crosslink I mean that both have links to each other, right now PR does not have a link to the discussion.

PR #1221, opened 2021-09-28
LdapSyncCommand: disable users missing in LDAP Active DN

vladsf:
LdapSyncCommand lacks an option to search and disable Eventum users which are not found in LDAP Active DN. Current code expects to find inactive users in Inactive DN.

This patch proposes --disable-missed option to check active Eventum users against LDAP. Users which are not found in Active DN to be set inactive.

### DIFF
--- a/src/Console/Command/LdapSyncCommand.php
+++ b/src/Console/Command/LdapSyncCommand.php
@@ -40,6 +40,7 @@ class LdapSyncCommand extends BaseCommand
         $this
             ->addOption('dry-run', null, InputOption::VALUE_NONE)
             ->addOption('create-users', null, InputOption::VALUE_NONE)
+            ->addOption('disable-missed', null, InputOption::VALUE_NONE)
             ->addOption('no-update', null, InputOption::VALUE_NONE)
             ->addOption('no-disable', null, InputOption::VALUE_NONE)
             ->addOption('no-notify', null, InputOption::VALUE_NONE);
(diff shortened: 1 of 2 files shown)


glensc:
please avoid those click-bait titles, unable to understand what the changeset is just by looking the title. so one must click on the diffs to get even a grasp of the proposed changes

vladsf:
@glensc Shall I expect this PR to be accepted?

glensc:
@vladsf there's no email notification once you mark PR ready for review, so you need to ask implicitly once it's ready, or just wait

glensc:
I don't understand the PR title nor changelog entry what this PR does. please improve. what is users missed or missed users?

src/Console/Command/LdapSyncCommand.php
```diff
@@ -40,6 +40,7 @@ protected function configure(): void
         $this
             ->addOption('dry-run', null, InputOption::VALUE_NONE)
             ->addOption('create-users', null, InputOption::VALUE_NONE)
+            ->addOption('disable-missed', null, InputOption::VALUE_NONE)
```
glensc, line 43: `--disable-missing` then perhaps?

src/Console/Command/LdapSyncCommand.php
```diff
@@ -98,6 +101,38 @@ private function updateUsers($enabled): void
         }
     }
 
+    /**
+     * Process active users missed from ldap
```
glensc, line 105: missing not missed

PR #1238, opened 2021-10-19
Support GitHub Actions for release build

glensc:
refs:
- https://github.com/ncipollo/release-action


glensc:
This `ncipollo/release-action@v1` action caused github support to revoke all my tokens and reset password due suspicious activity.

> 
> Hi glensc,
> 
> We’re writing to let you know that we observed suspicious activity that suggests a threat actor used a Personal Access Token (PAT) associated with your account to access private repository metadata.
> 
> Out of an abundance of caution, we reset your account password and revoked all of your Personal Access Tokens (classic), OAuth App tokens, and GitHub App tokens to protect your account, glensc.
> 
> * What happened *
> 
> We do not believe the attacker obtained these tokens via a compromise of GitHub or its systems, because the tokens in question are not stored by GitHub in their original, usable formats. At this time, evidence suggests that the threat actor used compromised tokens to access private repository metadata. You can review an example of repository metadata in the example response here:
> 
> https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-repositories-for-a-user--code-samples
> 
> The following tokens were identified:
> 
> automatic releases for eventum/eventum

Issue #1174, opened 2021-08-05
LDAP login failing on 3.10.X


areis422:
**Version(s) affected**: 3.10.x (specifically 3.10.5 and 3.10.6)
**PHP Version**: 7.4.22 (didn't change between 3.9.2 and 3.10.6)

**Description**  
With LDAP enabled and working in 3.9.2, upgrading to 3.10.X results in the following message after submitting the login:
`
Sorry, there was an error logging you in: Call to a member function get() on null
`

**How to reproduce**  
1. have instance running on 3.9.2 with LDAP via Active Directory enabled.
2. Upgrade to 3.10.6
3. Fail on login

**Additional context**  
Log messages:
```
[2021-08-05 01:16:33] auth.INFO: Login attempt by 'user' was successful. {"user":"user","type":"success","extra":null} {"file":"/var/www/eventum-3.10.6/lib/eventum/class.auth.php","line":85,"class":"Auth","function":"saveLoginAttempt","memory_peak_usage":"4 MB","memory_usage":"4 MB","url":"/login.php/login.php","ip":"x.x.x.x","http_method":"POST","server":"staging.eventum.domain.tld","referrer":"https://staging.eventum.domain.tld/login.php"}
[2021-08-05 01:16:33] app.CRITICAL: Unable to instantiate auth adapter {"exception":"[object] (Error(code: 0): Call to a member function get() on null at /var/www/eventum-3.10.6/src/ServiceContainer.php:89)"} {"file":"/var/www/eventum-3.10.6/lib/eventum/class.auth.php","line":476,"class":"Auth","function":"getAuthBackend","memory_peak_usage":"2 MB","memory_usage":"2 MB","url":"/list.php","ip":"x.x.x.x","http_method":"GET","server":"staging.eventum.domain.tld","referrer":"https://staging.eventum.domain.tld/login.php"}
```

Auth config (from setup.php):
```php
    'auth' => [
        'adapter' => 'Eventum\\Auth\\Adapter\\ChainAdapter',
        'options' => [
            'Eventum\\Auth\\Adapter\\MysqlAdapter' => null,
            'Eventum\\Auth\\Adapter\\ChainAdapter' => [
                [
                    'Eventum\\Auth\\Adapter\\MysqlAdapter',
                    'Eventum\\Auth\\Adapter\\LdapAdapter',
                ],
            ],
            'Eventum\\Auth\\Adapter\\LdapAdapter' => null,
        ],
        'login_backoff' => [
[...]

PR #1254, opened 2021-11-09
[3.11] Remove deprecated Eventum\Monolog\Logger class

glensc:


glensc:
Watch for the problem in https://github.com/eventum/eventum/issues/1174

PR #1309, opened 2022-01-10
Chore: Update smarty to 4.0.3

glensc:
Try trial error, unless there's a guide:
- https://github.com/smarty-php/smarty/pull/629#issuecomment-1008778679


glensc:
Stopped for now, as I run into 500 error on local development and no errors anywhere.

PR #1483, opened 2023-01-13
Various fixes for PHP 8 compatibility

inguin:
See individual commits for details.


glensc:
So now you just omit `escapeInteger` calls, making code less secure again?

vladsf:
> The 'sort_by' parameter is validated in saveSearchParams(), so there
should be no need for quoting.

What if `$sort_by` is later used unaware of being unquoted?

inguin:
> So now you just omit `escapeInteger` calls, making code less secure again?

The entire search profile is serialized upon save, so there should be no need for escaping at this point.

The individual values are still escaped in `buildWhereClause()`.

inguin:
> What if `$sort_by` is later used unaware of being unquoted?

I'm not sure what the concern is. `$sort_by` holds the name of a known database column (one of `SORT_BY_FIELDS`). It should be perfectly safe to be used without quoting.

vladsf:
May be `Misc::escapeInteger` shall be replace with `Misc::escapeId` which returns unsigned int or empty string?

vladsf:
> > What if `$sort_by` is later used unaware of being unquoted?
> 
> I'm not sure what the concern is. `$sort_by` holds the name of a known database column (one of `SORT_BY_FIELDS`). It should be perfectly safe to be used without quoting.

Not sure I understand. Right a few lines below `$sort_by` is used unescaped in SQL query:
[SQL query](https://github.com/eventum/eventum/blob/f795e9311178c52ed5cded9c9a73cc17efeae39a/lib/eventum/class.search.php#L381)

I guess this is unsafe.

inguin:
> Right a few lines below `$sort_by` is used unescaped in SQL query: [SQL query](https://github.com/eventum/eventum/blob/f795e9311178c52ed5cded9c9a73cc17efeae39a/lib/eventum/class.search.php#L381)
> 
> I guess this is unsafe.

No, it's not, because `saveSearchParams()` uses a whitelist to make sure that the only possible values are `last_action_date`, `pri_rank`, `iss_id`, `sta_rank`, `iss_summary`, or `custom_field`. (BTW, `custom_field` is a separate branch without any quoting even before my change).

The call to `quoteIdentifier()` was introduced on Nov 9 2021 (c732276e0855db9c0d3bafb33d41484ed34202ba).
The whitelist was introduced one day later (9119c50e86386a4942b6aaf099448df07ad97dbc), making the quoting redundant.

If you feel it's safer I can also change the regexp to deal with quoted identifiers.

vladsf:
`saveSearchParams()` method used to save the current search parameters in a cookie, on a user side. Cookies are editable by user no matter what whitelists were used.

inguin:
> Cookies are editable by user no matter what whitelists were used.

No cookies are involved as far as I know.

Search filters are taken from the request (`$_GET` or `$_POST`), or the search profile persisted in the database.

vladsf:
So ` $options['sort_by']` directly comes from GET/POST request? Is that correct?

inguin:
> So ` $options['sort_by']` directly comes from GET/POST request? Is that correct?

Correct, with validation against `SORT_BY_FIELDS`.

vladsf:
> 
Sorry I can not find it. Please point me the line where `$options['sort_by'] ` is sanitized or validated.

inguin:
> Sorry I can not find it. Please point me the line where `$options['sort_by'] ` is sanitized or validated.

[class.search.php, line 92](https://github.com/eventum/eventum/blob/master/lib/eventum/class.search.php#L92)

glensc:
🚧 Never link to branches, share a permalink 🚧
- see https://news.ycombinator.com/item?id=8046710
  and https://help.github.com/articles/getting-permanent-links-to-files/
  and https://docs.gitlab.com/ee/user/shortcuts.html#project-files
- hit `y`
- and perhaps for https://github.com/josephfrazier/octopermalinker (GitHub)

NOTE: edit your post and fix it with a permalink!

PS: gitlab has a bug and you need to reload a page before `y` works.

inguin:
> NOTE: edit your post and fix it with a permalink!

Done.

vladsf:
> > Sorry I can not find it. Please point me the line where `$options['sort_by'] ` is sanitized or validated.
> 
> [class.search.php, line 92](https://github.com/eventum/eventum/blob/8b71ab5fb20796233ae7731354d3b49ce21c8458/lib/eventum/class.search.php#L92)

I can see now. Yet this is so inconsistent and might hit once day later.

glensc:
Is this ready for merge? please rebase against `origin/master`. I won't do deep review, but looks ok.

glensc:
please also modify composer.json to allow php versions you tested. don't don't forget to update composer.lock

inguin:
> please also modify composer.json to allow php versions you tested.

I'm testing with PHP 8.1.14 (Fedora 37) and PHP 8.2.1 (CentOS 7.9). `composer.json` requires `^8.0`, which should cover both.

glensc:
they say that php engine versions aren't following semver, so ~8.0, ~8.1, ~8.2 would be more accurate to say, so it wouldn't include 8.3

inguin:
> they say that php engine versions aren't following semver, so ~8.0, ~8.1, ~8.2 would be more accurate to say, so it wouldn't include 8.3

OK, makes sense. Done.

glensc:
Your lock update is wrong, you ran `composer update`, updating all dependencies, I asked to run `composer update --lock`.

Doesn't make sense to put changelog and composer.json change to same commit, totally different changes. not atomic commit. now if I want to revert the composer.json commit I will also revert changelog change as it's same commit. that's about atomic commits principle.
- https://www.freshconsulting.com/insights/blog/atomic-commits/

and "update composer.json" is a clickbait commit title. have to git show to see what the change is about :) "Allow PHP 8.0-8.2 PHP versions in composer" would be a more appropriate title.
- https://cbea.ms/git-commit/

glensc:
Oh, I didn't realize "^8.0" is already there from 2019:
- https://github.com/eventum/eventum/commit/6c3c63c88e49a9a19a854ce1dffc7e87d7d88e0d

inguin:
> Oh, I didn't realize "^8.0" is already there from 2019

I thought that why you asked to make it more specific. :grin:

Want me to drop the commit?

glensc:
> I thought that why you asked to make it more specific. 😁

yes, we wanted and you agreed. so do it. but properly this time :)

PR #1486, opened 2023-01-23
Fix "chosen" select box rendering in Firefox

inguin:
Select boxes enriched with jquery-chosen are not properly rendered in Firefox: The dropdown is permanently visible (albeit initially empty), even if the select box is closed.

The issue seems to be caused by custom CSS rules which date back to a time when the plugin worked differently.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ See [Upgrading] for details on how to upgrade.
 
 - Fix exception in ManageBaseController, LDAP login, #1435, #1482, #1174
 - Various fixes for PHP 8 compatibility, #1483
+- Fix "chosen" select box rendering in Firefox, #1486
 
 [3.10.13]: https://github.com/eventum/eventum/compare/v3.10.12...master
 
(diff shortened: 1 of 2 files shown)


glensc:
Screenshots?

inguin:
> And this will continue to work on Chrome?

Yes. So far I've tested Firefox, Chrome and Chromium.

Generally I'd assume that the jquery-chosen plugin has pretty good browser compatibility by default. Can you recall what prompted the addition of those custom rules back in 2015?

glensc:
I suppose you ran git blame and found commits for those changes. your guess is as good as mine. I only know I'm not a CSS guru, so the CSS was likely provided by somebody else

the position relative came from initial integration:
- https://github.com/eventum/eventum/commit/46366d2088b0df69446db45ddd93b115fff2e3b0 (https://github.com/eventum/eventum/pull/12/files)

inguin:
> I only know I'm not a CSS guru

Neither am I. I only saw that the [jQuery Chosen demo page](https://harvesthq.github.io/chosen/) works flawlessly in all my browsers, so I disabled Eventum's custom CSS rules one by one until it looked fine there as well.

The Chosen plugin was upgraded from 1.2.0 to 1.8.7 in #812, so my best guess is that whatever the rules were supposed to fix has also been addressed in the original project by now.

PR #1537, opened 2024-08-08
Fix rendering of pie charts with PHP 8

inguin:
Eventum uses an ancient version of PHPlot which uses the old-style constructor syntax that is no longer supported by PHP 8. As a result the pie charts on the project home page fail to render.

To work around the issue, call the constructor manually.

### DIFF
--- a/src/Controller/Helper/PlotHelper.php
+++ b/src/Controller/Helper/PlotHelper.php
@@ -46,6 +46,10 @@ class PlotHelper
     private function create($width, $height)
     {
         $plot = new PHPlot($width, $height);
+        if (PHP_MAJOR_VERSION >= 8) {
+            // Hack to make PHPlot 6.1 work with PHP 8: manually call the old-style constructor
+            $plot->PHPlot($width, $height);
+        }
         $plot->SetTTFPath($this->fonts_path);
         $plot->SetUseTTF(true);
 


glensc:
Why not upgrade phplot?

glensc:
Can you add changelog entry commit?

inguin:
> Why not upgrade phplot?

That works, too. I tried that as well, it only needs a few small changes.

I'll give it a try if you'd prefer that.

> Can you add changelog entry commit?

Sure, will do.

inguin:
I've updated the PR to upgrade phplot instead.

The branch name "phplot-workaround" is no longer perfectly appropriate, but apparently I can't change it in an existing PR.

glensc:
Their composer.json specifies no version constructs:
- https://github.com/PHPlot/phplot/blob/e36d9513e1e0df5d75784e98e6ca4ae745ba27c9/composer.json#L6

so, I'll assume it works with PHP version range eventum currently supports.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 See [Upgrading] for details on how to upgrade.
 
+- Update PHPlot to v8.0.0.3, #1537
```
glensc, line 7:
```suggestion
- Update PHPlot to v8.0.0.3 to fix rendering with PHP 8, #1537
```

